PR #166, opened 2015-02-02
Phases 1-3 of libretro-fetch.sh rewrite

iKarith:
This PR makes libretro-fetch.sh work like libretro-build.sh in that you can fetch one or fetch all.  It's a lot _longer_ because I didn't want to change too much at once (see the commit to fix the stupid mistake regarding snes9x_next for my justification for that…)

Where this CAN go from here is to replace the fetch_project type macros with direct calls of the things they call.  Of course if you do that, you'll want to do something more clever with how the echo commands to tell you what it's currently doing are progressing.  I've got some thoughts for that too, but I want to try a couple of things to see what works.

Ultimately this benefits the OS X buildbot I'm working on which will fetch things as it needs to build them.  In radius's buildbot he has his own separate functions for doing that, but rather than duplicate the buildsystem I'd like to simply make it more flexible.  Fetching seemed like a good place to start.


iKarith:
As of now, you can:

tjcarter@amaya:~/Source/libretro-super$ ./libretro-fetch.sh fetch_libretro_nestopia fetch_libretro_bsnes_mercury       
tjcarter@amaya:~/Source/libretro-super$ ./libretro-build.sh build_libretro_nestopia build_libretro_bsnes_mercury       

…and it works as you might expect.  :)

PR #168, opened 2015-02-02
And another ${WORKDIR} bug *sigh*

iKarith:


iKarith:
I missed a ${WORKDIR} in a test for whether to do a clone or a pull, which can be observed to break ppsspp if something else gets updated before it does.  It only affects a couple of cores, so it was easy to overlook.

PR #169, opened 2015-02-03
Some build script refactoring

iKarith:


iKarith:
Also, this PR allows libretro-build.sh and libretro-fetch.sh to operate outside of the libretro-super dir, which is nice for keeping that directory clean (and also for use in buildbot scripts and the like. Note retroarch-build.sh IS NOT benefitted by these changes yet because it doesn't work on OSX, which means I haven't done anything with it yet.

PR #191, opened 2015-02-19
iOS build cleanup (use platform=ios ./libretro-build.sh now)

iKarith:
- Move iOS config to libretro-config.sh
- iOS build now happens using more sensible platform=ios
- Documentation to explain what/why


iKarith:
My mame rule refactor got added to this.  The changes are fairly trivial, but haven't been FULLY tested because mame does not currently compile.

PR #197, opened 2015-02-21
Should get QNX, PS3, and PSP1 compiling…

iKarith:
This is an UGLY fix that simply passes CC and CXX to make only if they're both set, everywhere make is called in libretro-build-common.sh.  It also makes sure the QNX, PS3, and PSP1 scripts all set WORKDIR safely.

The changes to libretro-build-common.sh should be reverted at the first opportunity, either after 1.1 is safely shipped or after all of the scripts that depend upon not having CC and CXX set are fixed or merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,8 +61,10 @@
 /libretro-vecx/
 /libretro-virtualjaguar/
 /libretro-yabause/
+/libretro-fuse/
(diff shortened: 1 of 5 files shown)


iKarith:
Actually "both" is not at all true.  It passes both if either are set.

PR #221, opened 2015-03-03
SOME color in the scripts, can be disabled in libretro-config{,-user}.sh

iKarith:
An admonition: Just because you CAN use colors in the scripts now doesn't mean you always SHOULD.  If I find someone's made these scripts look like an angry fruit salad, I'll be reverting those changes as I already told Twinaphex I would.  Too much color makes the script output loud and distracting, less is more, etc.

Nutshell: As it stands, I consider libretro-fetch.sh to be too loud already.  The only reason it does any color at all is because at some point libretro-build.sh and libretro-fetch.sh will be special cases of libretro-super that (just) build and (just) fetch.  The normal thing will be to fetch, compile, and probably package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,6 @@
 /libretro-mame/
 /libretro-mame078/
 /libretro-mame139/
-/libretro-manifest/
 /libretro-mednafen_gba/
 /libretro-mednafen_lynx/
 /libretro-mednafen_ngp/
@@ -62,9 +61,9 @@
 /libretro-virtualjaguar/
 /libretro-yabause/
 /libretro-fuse/
-/libretro-dat-pull/
-/libretro-common/
-/libretrodb/
+/libretro-lutro/
+/libretro-gw/
+/libretro-devkit/
 /retroarch/
 /build-summary.log
 /build-revisions/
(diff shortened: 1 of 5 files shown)


twinaphex:
These colors don't work on OSX. They display a bunch of nonsensical characters instead.

Change it so that colors are entirely optional instead of being the default.

iKarith:
OSX is where I tested them. Will change the default, though.

Though which OSX did you test them on that they don't work? Every version going back to at least 10.2 has had this much.

Example of weirdness?

Joseph
Sent via mobile

iKarith:
I wonder if older versions of Terminal don't like ^[[0;0m maybe? That's legal (and convenient), but somewhat nonsensical to reset the attribute twice.

Joseph
Sent via mobile

twinaphex:
This is on Yosemite though where I tested it on.

iKarith:
Apple's bash seems to disable character substitution on purpose. Neither echo -e nor printf (shell builtin) do what they are spec'd to do sinc at least bash 2.0.

In my own personal scripts (which run on a couple of non-bash shells now and then), I insert a literal escape character by typing ctrl-v escape in vin. I figured sticking non-7bit printable characters in a script wasn't best practice, but if Apple's gonna be stupid, we'll play along.

Joseph
Sent via mobile

PR #227, opened 2015-03-06
Fix to libretro-fetch.sh

almostalive:
@twinaphex this fixes libretro-fetch.sh the register_core line needs to use _ instead of - in the names.

@iKarith something is wrong with libretro-build.sh. I think it has to do with https://github.com/libretro/libretro-super/commit/c5dba10034050048d166947e2b813fb11e68c84d but to complicated for me to figure out.

### DIFF
--- a/libretro-build.sh
+++ b/libretro-build.sh
@@ -169,8 +169,8 @@ else
 	if [ $FORMAT_COMPILER_TARGET != "ios" ]; then
 		# These don't currently build on iOS
 		build_libretro_bnes
-		build_libretro_core ffmpeg
-		build_libretro_core ppsspp
+		libretro_build_core ffmpeg
+		libretro_build_core ppsspp
 	fi
 	libretro_build_core o2em
 	libretro_build_core hatari
--- a/rules.d/core-rules.sh
+++ b/rules.d/core-rules.sh
@@ -328,23 +328,23 @@ libretro_lutro_name="Lutro"
 libretro_lutro_git_url="https://github.com/libretro/libretro-lutro.git"
 libretro_lutro_build_makefile="Makefile"
 
-register_core "lutro-platformer"
+register_core "lutro_platformer"
 libretro_lutro_platformer_name="Lutro-Platformer"
 libretro_lutro_platformer_git_url="https://github.com/libretro/lutro-platformer.git"
 
-register_core "lutro-tetris"
+register_core "lutro_tetris"
 libretro_lutro_tetris_name="Lutro-tetris"
 libretro_lutro_tetris_git_url="https://github.com/libretro/lutro-tetris.git"
 
-register_core "lutro-snake"
+register_core "lutro_snake"
 libretro_lutro_snake_name="Lutro-snake"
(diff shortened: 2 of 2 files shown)


almostalive:
well I meant for the change to libretro-build.sh to be a separate pull request but github threw it into here. To further clarify about the libretro-build.sh problem, it seems to only build cores with special rules and skips all the normal cores. The bsnes's and mame type stuff will build, but not the others.

twinaphex:
Can you do a git pull right now without your commits and see if this issue still remains?

I think everything works as expected again.

almostalive:
=== Lutro
Fetching lutro...
cd "/tmp/makepkg/libretro-super/src/build-libretro/libretro-lutro"
git pull
Already up-to-date.
=== -platformer_name
libretro_fetch:Unknown fetch rule for lutro-platformer: "-platformer_fetch_rule"

with my changes to libretro-fetch.sh it fixes that problem. and the commit to libretro-build.sh gets ppsspp and ffmpeg building again because those two items didn't get changed. github is pissing me off so if you want you can close this and do the changes yourself.

twinaphex:
I'll just merge it. Sorry for the downtime these changes have caused in making things build again with libretro-super. This kind of breakage will no longer be considered acceptable and from now on I will be regression-testing every commit that gets made to libretro-super because we can't really have our build system be broken for days on end and then when I ask the persons involved in the changes to please comment days go by with no response.

iKarith:
What is your libretto-fetch command line? It looks like you're passing something extra.

If you need a patch in there for makepkg, feel free to add it.

Joseph
Sent via mobile

almostalive:
my libretro-fetch line is just ./libretro-fetch.sh
this is the PKGBUILD i use for building all the cores https://github.com/almostalive/stuff/blob/master/libretro-super/PKGBUILD I've been working on that for a long time. Nothing to much changes other than adding new cores to sources. I have makepkg set to save the sources so i always start with a clean slate and don't have to keep downloading them over and over. there is also a libretro-super-git PKGBUILD on Arch's AUR https://aur.archlinux.org/packages/libretro-super-git/ with this pkgbuild every time  you run it, you have to download everything again which sucks. I use to maintain must of the individual cores that are on aur so I could always have a clean saved source but i got tired of that and made my personal libretro-super pkgbuild

in my bad testing i didn't run libretro-fetch on it own, should of done that, but the changes i did to register_core names make sense, since all the folders are named libretro-something_if_it_has_spaces and i'm figuring you get the folder names from that now. (could be wrong just a guess). when the lutro games had - instead of _ makepkg would fail right after that line i posted earlier. my guess why was because fetch was expecting _ in the names and not -, changing it fixed it so i sent a pr.

my change to libretro-build.sh was just something that got missed and at the time I made this pr, build was messed up but you got that fixed right away.

so right now everything is running great. Your work on libretro-super has made it much more complicated but has also made it much nicer which is nice. Hell I've been thinking about an addition to your work that might cut down on some scripts but i need time to look it over and write my idea in a very understandable way for you to do that work since my bash skills are very poor but I have enough to get my by.

PR #238, opened 2015-03-12
More modern rules support!

iKarith:
Rules may now:

* Specify extra arguments to make
* Define more than one core produced by the makefile
* Specify that cores are placed in a directory other than the one with the makefile

This means bnes, bsnes, bsnes_cplusplus98, bsnes_mercury, and emux are no longer legacy.

mame, pcsx_rearmed, and fba (sort of) still are.  I can look at fba later, but mame needs to build on ios again before I can do anything with it.


iKarith:
That per-core shell function works like this:

```
<pre-build function>
<pre-clean function>
make clean
<post-clean function>
make
<post-build function>
copy
```

I'm not sold on that.  Eventually what I want to do is have it so fetch, clean, build, and package are separate functions done to a core.  I've found myself wanting to have a "make clean for this core, however you do that" command.  So I might change this a bit.  Input welcome at this stage as the only thing that uses any of it is my early-stages modern mupen rule.

iKarith:
Solution detected:

```bash
$core_build_configure
$core_build_preclean
$core_build_prebuild
$core_build_prepkg
```

With the last one being named with the assumption that the copy function will become a packaging function at some point.

PR #250, opened 2015-05-09
Update local fork

ProfessorKaos64:
update fork


ProfessorKaos64:
Disregard, wrong direction.

PR #271, opened 2016-01-15
Add PUAE to Android buildbot cores recipes

diablodiab:


diablodiab:
I've made a new pull request to the UAE core which should fix this build error. There might be other errors after this, as the source apparently uses some deprecated functions which cause problems on newer Android API versions. If it still fails to compile please give me a screenshot and I'll look into it.

PR #283, opened 2016-04-30
Use autoconfig everywhere instead of autoconf

Kivutar:


fr500:
That script is only for Windows and it should not affect the issue at hand
at all

PR #285, opened 2016-05-18
Configure new build architecture "armeabi-v7a-neon"

meepingsnesroms:


fr500:
hmmm that isn't used at all, except for local scripts that you or
@Twinaphex may use, the buildbot just uses libretro-buildbot-recipe.sh

fr500:
Also, should we just enable neon for the armv7a abi altogether.
The downloader won't know which ones to pick so this doesn't really add anything.

What is the name of a compiled core with this rule?

meepingsnesroms:
If your compiling for neon set V7NEONOPTIMIZATION="1" to tell the jni makefile to use neon,all other optimizations are in the jni makefile and will be applied regardless.

I modified this file so that armv7neon would be considered its own architecture,so file names remain the same but the directory is armeabi-v7a-neon instead of armeabi-v7a,so retroarch can detect if the device is neon and set the core download path.

meepingsnesroms:
Don't merge neon with regular armv7 that will break some devices.(Tegra 2)

fr500:
hmmm I see no easy way to solve this, I assume the optimizations would be desirable for RetroArch too?
Maybe we could have them in the same folder with a _neon suffix instead but it would add to the confussion

Not sure what to do here.

meepingsnesroms:
This would be good for retroarch too but would have to be in a separate apk because you cant for neon check before downloading it.(the user would have to check)

You can check if the device has neon,and if it does set the download directory to neon.(cpufeatures.h for android)

fr500:
Swapping the url is easy, the build process is rather complicated so I assume one of the folders would have less cores or something.
There are several toolchains at play, and different build commands depending on the toolchain in use.

Do they provide a meaningful boost?
Are you sure we would need a different retroarch for neon devices? that seems like a no-go, not sure what @twinaphex thinkgs.

meepingsnesroms:
Yes,if you have a for loop checking an array of colors and copying from a different buffer if the top bit is set(transparent bit) then you could do 5 pixels all at once with neon instead of check and copy 5 times in a for loop.

twinaphex:
I'll keep an open mind to this but I'll have to return to this within a few days.

twinaphex:
Guess I'll just merge this. If it poses problems we will return to it later.

PR #286, opened 2016-05-18
Add fb_alpha_new to Android build list

diablodiab:
Use the new JNI files

### DIFF
--- a/libretro-build-android-mk.sh
+++ b/libretro-build-android-mk.sh
@@ -134,6 +134,7 @@ WANT_CORES=" \
 	dinothawr \
 	desmume \
 	fb_alpha \
+	fb_alpha_new \
 	bsnes_mercury_performance \
 	bsnes_performance \
 	mame2000 \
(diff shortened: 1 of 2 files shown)


fr500:
hmmm actually these aren't used at all

diablodiab:
libretro-build-android-mk.sh is not the script that contains the list of Android cores being built by the buildbot ?

diablodiab:
Oh I think I see. It's cores-android-jni, right? Under recipes for android ?

fr500:
no, for jni cores it would be this:
https://github.com/libretro/libretro-super/blob/master/recipes/android/cores-android-jni

Example:
fb_alpha libretro-fba https://github.com/libretro/fba-libretro.git PROJECT YES GENERIC_JNI makefile.libretro svn-current/trunk/projectfiles/libretro-android/jni 

CORE_NAME FOLDER URL PROJECT|SUBMODULE|BRANCHE ENABLE JOB_TYPE MAKEFILE_NAME MAKIEFILE_LOCATION

So, important things here:
- CORE_NAME is the name of the build product, in this case fba_libretro
- FOLDER just the folder for the local clone, libretro-fba_new should do the trick
- URL
- PROJECT most of the time
- ENABLE would be YES
- GENERIC_JNI

The rest should be self explanatory, just make sure to include a trailing space always.
If you have problems I can do it but it's good to get other people involved.

diablodiab:
Does it still make sense to keep the changes I made to libretro-build-android-mk.sh ?

diablodiab:
Updated cores-android-jni as well now. I assume that it is okay to have libretro-build-android-mk.sh updated so you can compile it that way also.

fr500:
yeah it would be fine I guess

PR #306, opened 2016-07-22
add hack to copy ParaLLEl lib

hizzlekizzle:
How does this look?


Alcaro:
Isn't it missing a 'then'? I can never remember how bash if statements work.

And you could improve the indentation.

PR #324, opened 2016-08-21
The main DAT must be at the end, see libretro/RetroArch#2935.

sergiobenrocha2:


RobLoach:
This will fix old titles being in the `metadat` folders.

PR #340, opened 2016-09-13
Fix misnamed info files

orbea:
I screwed the names for these accidentally... :/


orbea:
Here is bit more of an explanation.

When submitting these info files earlier I made two mistakes.
* There should be `-libretro` appended to them to match the core and other info files.
* I typo'd `Cruzes` as `Cruze`.

PR #341, opened 2016-09-13
Switch Serial Scanning to CRC checks

RobLoach:
This is in response to https://github.com/libretro/RetroArch/pull/3570 .


RobLoach:
Going to close this until we decide to make this change.

PR #342, opened 2016-09-14
Add glupen64 for android

loganmc10:
This allows:
```
./libretro-build-android-mk.sh glupen64
```
to work. The result ends up at ```./dist/android/armeabi-v7a/glupen64_libretro_android.so```

I'm not sure if there is anything else that needs to be changed in order for the core to get zipped up or whatnot.

In order for this to work the buildbot needs a couple things: **flex** and **bison** need to be installed if they are not.

The buildbot needs the Android standalone toolchain installed.

Basically:
./android-ndk-path/build/tools/make_standalone_toolchain.py --arch arm --install-dir ~/mychain

PATH=$PATH:~/mychain/bin   (arm-linux-androideabi-gcc needs to be in the PATH somewhere)

### DIFF
--- a/libretro-build-android-mk.sh
+++ b/libretro-build-android-mk.sh
@@ -39,6 +39,22 @@ die()
 	#exit 1
 }
 
+build_libretro_standalone()
+{
+	cd $BASE_DIR
+	if [ -d "libretro-${1}" ]; then
+		echo "=== Building ${1} ==="
+		cd libretro-${1}
+		if [ -z "${NOCLEAN}" ]; then
+			platform=android make clean
+		fi
+		platform=android make -j4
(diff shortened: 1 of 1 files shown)


twinaphex:
@fr500 Can you review this?

fr500:
I already added a glupen target to the buildbot, this is for super.

twinaphex:
Does glupen64 require a separate rule, it can't use the existing standalone toolchain rule that already exists in libretro-super?

fr500:
I have no idea, I don't use super on the buildbot

loganmc10:
I guess if the buildbot doesn't need this it isn't needed, I thought the buildbot used this

PR #347, opened 2016-09-24
Fix Microsoft - MSX 2 naming

RobLoach:
No-Intro has renamed to `Microsoft - MSX 2`. We can now fix it here. I'll push up a fix to libretro-database.


RobLoach:
Actually, we actually want MSX2.

PR #353, opened 2016-10-06
Ability to configure cores to fetch/compile/install.

Wintermute0110:
This patch creates a configuration file so user can configure what cores to clone/build/install in an easy and convenient way. This configuration file will change much less than files in rules.d/*.sh, so user configuration will be maintained if libretro-super repository is update.

Patch contains minimum change to existing libretro-super files and I think many people will like it.


Wintermute0110:
Hi. Any comments on this?

PR #362, opened 2016-11-02
Add .info file for the Dolphin Launcher core

RobLoach:
The [Dolphin Launcher libretro core](https://github.com/RobLoach/libretro-dolphin-launcher) allows launching [Dolphin](https://dolphin-emu.org) from RetroArch. Giving us the ability to scan/browse Nintendo GameCube and Wii games from the RetroArch menu.

screenshot at 2016-11-02 07-52-46


twinaphex:
I'll tolerate this for now as long as we don't have a Dolphin libretro core. As soon as it is available, though, this will be killed with fire. 

I really hope this is the first and last time something like this is going to be proposed. The entire aim here is making libretro cores, not using RetroArch as a launcher for other standalone emus.

RobLoach:
> As soon as it is available, though, this will be killed with fire.

Agreed. Having native cores is a much better solution. :+1:

PR #383, opened 2016-12-12
Info update

garbear:
Recently @fetzerch created some infrastructure to automate generating the Kodi "fluff" for libretro cores. Looking at some of the diffs it created, I noticed some outdated info in libretro-super.

There is probably more to do, but I figured I'd PR the errors that I immediately noticed from the diffs.


twinaphex:
Split this up into several different PRs, I am already seeing some parts I don't want to merge in here, while the other parts are OK.

dist/info/mednafen_psx_libretro.info
```diff
@@ -1,23 +0,0 @@
-display_name = "PlayStation (Mednafen PSX)"
```
fr500: why are you removing this file?

garbear:
@twinaphex can you comment on the distasteful parts here in this PR? I don't know how these should be separated other than 10 individual PRs.

dist/info/mednafen_psx_libretro.info
fr500: The library is mednafen_psx_libretro.so|dll|dylib

twinaphex:
You're removing a mednafen psx core info file for no reason, I don't think all the Beetle/Mednafen renames are necessary or even desirable.

Honestly, just split it up into 10 separate PRs. I don't feel comfortable merging this all as one package.

dist/info/mednafen_psx_libretro.info
fr500: in any case the beetle one should go.

garbear:
k, it's been a long day so i'll try to get to this tmrw

fr500:
I think we don't want to go thru with completely renaming mednafen to beetle. The author was annoyed about use forking... but she might get more annoyed if it's completely renamed.

PR #388, opened 2016-12-13
Update FCEUmm version from changelog.txt

garbear:
See https://github.com/libretro/libretro-fceumm/blob/master/changelog.txt


garbear:
oops, accidentally deleted branch. not sure if you want this change, I just figured last known FCEUmm version was desirable over "SVN", which doesn't seem to apply to a git repo

RobLoach:
Seems like a good change.

sergiobenrocha2:
It's not really v0.98.13, maybe v0.98.13+ or v0.98.13+SVN

http://cah4e3.shedevr.org.ru/fceultra.php

garbear:
I know less about the history of this emulator, so the version may need additional decorations like pluses and tags. Feel free to add them. I just wanted to get the numbers right.

PR #393, opened 2016-12-13
Update Beetle info files

garbear:
Apologies for not knowing the politics behind the rename, but I noticed that all github repos were renamed to beetle, so I figured to update the display names and .info names. Currently there is a mix of mednafen and beetle files, so I made them all consistent.

The mednafen psx file has been removed because an identical beetle psx file exists, minus the name change. Does libretro build two cores from the same repo with different names?


dist/info/beetle_bsnes_libretro.info
```diff
@@ -1,7 +1,7 @@
-display_name = "SNES / Super Famicom (Mednafen bsnes)"
+display_name = "SNES / Super Famicom (Beetle bSNES)"
 authors = "byuu|Ryphecha"
 supported_extensions = "smc|fig|bs|st|sfc"
-corename = "Mednafen/Beetle bsnes"
+corename = "Beetle bSNES"
```
Alcaro: https://gitlab.com/higan/higan/blob/72b6a8b32e73c27346f96327e15e15e9e00357f5/higan/sfc/sfc.hpp#L13 says it is not called bSNES.

garbear: While researching emulator histories for Kodi's add-on descriptions, one of the oldest bsnes references I came across stylized it this way. Beetle is a fork of a much more recent version though, so I agree with you that "Beetle bsnes" is fine.

Speaking of which, I'm trying to write brief descriptions for all the cores so we can show stuff in our GUI and also because emulator history fascinates me. For example, two of my descriptions:

Beetle bsnes:
> Beetle bSNES is the SNES module of Beetle, a fork of Mednafen. Mednafen's SNES module itself is a fork of bSNES v0.59.

bSNES-Mercury:
> bSNES Mercury is a fork of Higan (originally called bSNES), notable for emulating the original hardware as accurately as possible through low-level, cycle-accurate emulation.
>
> The "Accuracy" profile focuses on accuracy at the cost of performance.


Would you guys be interested if I shared these descriptions with your .info files?

garbear:
Any opinions on making these core names consistent?

Also, @fetzerch and I have a script that generates our add-ons and also creates a summary on our wiki: http://kodi.wiki/view/Game_add-ons. This shows both info extracted from the core and from .info files of libretro-super. You can look over this to see if everything is correct.

I plan to update the script to update the .info files as well so that your repo can stay in sync. Expect some PRs if there's any updates.

RobLoach:
Seems like a good change.

fr500:
it wouldn't work without changing the build product names...
ie, the new filenames won't match the library names.

garbear:
I didn't know they were connected. It seemed arbitrary, because PS had both a madnafen and beetle .info file.

I can send PRs upstream if it would be welcome.

twinaphex:
Upstream? Hah, that's a joke with Mednafen. Seriously, he/she would be offended by the mere sight of a core info file, he/she does not even have a repository and still lives in the '90s with an outdated 'sneakernet' style approach to releasing builds. I think for now we can close this PR.

garbear:
I mean to the libretro repos... "side stream" would have been a better term I guess?

PR #396, opened 2016-12-24
Update libretro-buildbot-recipe.sh

fr500:


fr500:
this should fix the redist... hopefully

PR #416, opened 2017-01-07
add Videopac+ BIOS checksums

markwkidd:
Currently only the US (Magnavox Oddessy2) BIOS is listed. This PR adds the three European Videopac+ BIOSes along with checksums taken from the versions recommended by René van den Enden on their website.


dist/info/o2em_libretro.info
```diff
@@ -1,4 +1,4 @@
-display_name = "Odyssey2 / Videopac+ (O2EM)"
+edisplay_name = "Odyssey2 / Videopac+ (O2EM)"
```
RobLoach: `edisplay_name` should be `display_name`

markwkidd:
Thanks @RobLoach! I just pushed a fix for the display name as suggested.

PR #420, opened 2017-01-09
update PC-FX BIOS filename and description

markwkidd:
Updated required BIOS filename per this PR to the libretro core: https://github.com/libretro/beetle-pcfx-libretro/pull/15

Also adds more detail about which version of the BIOS is required.


RobLoach:
:+1:

PR #434, opened 2017-01-22
(info) Update vecx license to GPLv3

orbea:
I'm not sure where @Kivutar & @sergiobenrocha2 got GPLv2 or LGPLv2.1 from, but as far as I can tell vecx is GPLv3.

See here. https://github.com/jhawthorn/vecx/commit/be44a67bcd7a8d71b631e1a53fcbf3167344085e

Also see the two PRs that added the older license information that this PR would replace.
https://github.com/libretro/libretro-super/pull/74
https://github.com/libretro/libretro-super/pull/142


sergiobenrocha2:
It's LGPLv3: https://twitter.com/nzeemin/status/521912237146251264

orbea:
Thanks for sourcing that, I guess the license is inconsistent. I don't really care which one is used as long as its free. We should probably all be on the same page though. :)

orbea:
He additionally says GPLv3 here.

https://github.com/jhawthorn/vecx/issues/2

Issue #436, opened 2017-01-28
Move Android recipes for mame2016/2014 from ndk13 to old ndk10


r-type:
Seem there is some issues with ndk13 and mame2016/2014 android core.
For example it segfault if you enter dipswitch menu and it doesn't happen with other arch like lin64/win64

But if you build both core with ndk10 (I've tested ith ndk10d) , it doesn't segfault in dipswitch menu.

I don't know if you always have an old ndk setup with buildbot, but if it's the case could you look at to add recipe for mame2014/2016 using ndk10 instead of ndk13.

I built the both core using this config for ndk10d

```
export ANDROID_NDK_ARM=/opt/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64
export ANDROID_NDK_ROOT=/opt/ndk
export ANDROID_NDK_LLVM=/opt/ndk/toolchains/llvm-3.5/prebuilt/linux-x86_64
```
and makefile arg for mame2014
```
make -f Makefile platform=android -j10 TARGET=mame
```
and makefile arg for mame2016
```
make  OSD=retro RETRO=1 NOWERROR=1 NOASM=1 gcc=android-arm gcc_version=4.9 TARGETOS=android-arm CONFIG=libretro NO_USE_MIDI=1 OS=linux VERBOSE=1 verbose=1 TARGET=mame -j10
```
@fr500 @twinaphex  if you can look at this ,i would be great!

r-type:
we can add 2 mame recipes for old ndk like this

cores-android-armv7-ndk-old-mame.conf
```
PATH home/buildbot/tools/android/android-ndk-r10e/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/
ANDROID_NDK_ARM /home/buildbot/tools/android/android-ndk-r10e/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/
ANDROID_NDK_ROOT /home/buildbot/tools/android/android-ndk-r10e/
ANDROID_NDK_LLVM /home/buildbot/tools/android/android-ndk-r10e/toolchains/llvm-3.5/prebuilt/linux-x86_64/
PLATFORM android
platform android
LIBSUFFIX _android
MAKE make
DIST armeabi-v7a
CORE_JOB YES
```
cores-android-armv7-ndk-old-mame
```
[...]

PR #437, opened 2017-01-29
move mame2014/2016 cores from ndk13b to  ndk10e

r-type:
I hope the format is correct  and that it will fine with 10e (works for me with 10d).
this should solve the dipswitch menu crash.


r-type:
https://github.com/libretro/libretro-super/issues/436

PR #443, opened 2017-02-16
Add meta hacks dats to db builds

JuanVCS:
Keeping it coordinated with https://github.com/libretro/libretro-database/pull/417


RobLoach:
This will help in organising some custom romhacks.

PR #445, opened 2017-02-18
Fix Nintendo DS (Download Play) to no longer be in BETA

RobLoach:
No-Intro no longer considers this DAT beta.


twinaphex:
You might need to rebase this.

RobLoach:
@twinaphex Thanks, fixed.

PR #471, opened 2017-06-17
Update MAME DBs for targeted MAME versions

RobLoach:
This will have MAME find games specific to the core version, with work from @markwkidd.

@markwkidd Mind having a look at this? What do you think is missing to get this in?


markwkidd:
Yes, all of the correlating MAME version numbers in the info files in this PR are correct. I know you've been putting some time into the MAME DATs here lately -- welcome to the madness of MAME's ROM system and thank you!

markwkidd:
@twinaphex would you consider merging this PR?

I believe there are many folks in addition to myself who are eager to see if this approach will be sufficient to address the challenges of MAME ROM scanning, and who will test this update to the databases.

twinaphex:
Alright, I'll assume RobLoach signed off on this and that this is good to merge.

PR #491, opened 2017-07-23
Change puae to jni makefile for x86 compatibility

meepingsnesroms:


meepingsnesroms:
I have checked the buildbot and android x86 is still not there, do changes in libretro-super require the buildbot to be taken offline to change, because when I change cores they are built within an hour.

PR #496, opened 2017-07-29
Add CHD to Supported Extensions for Redream

Awakened0:
CHD support was added: https://github.com/inolen/redream/commit/6b1345795d68f242c2184bcf174c8221cd26cf9e


RobLoach:
Thanks a lot! :+1:

PR #510, opened 2017-08-07
Update wiiu

r-type:
add puae


twinaphex:
Some compilation issues -

http://p.0bl.net/16077

r-type:
I added the files , but they should really be in /home/buildbot/tools/devkitpro/portlibs/ppc/include
it's strange buildbot doesn't have the portlibs include at least zlib one.

PR #534, opened 2017-09-09
Add NP2Kai to recipes. (Please Verify)

Tatsuya79:
1st time I'm doing that, please check if OK.


Tatsuya79:
I see he's using partly uppercased "NP2kai" as target... it's not going to work right?
Should I change that in the recipes or send a PR to him to make everything lowercase?

PR #542, opened 2017-09-18
[Redream] Correct firmware file names listed

alondero:
Redream looks for bios files called "boot.bin" and "flash.bin", not "dc_boot.bin" or "dc_flash.bin" (which Reicast expects). Updating the info file with the correct files.
Alternatively @inolen may want to consider supporting both name conventions to make it easier to interop between the two DC cores?


orbea:
Can this get merged?

inolen:
Had someone message me on IRC asking if this change looks correct, just confirming that it does.

PR #547, opened 2017-09-24
Update SameBoy, VBA-Next and Emux GB core info files

esoptron:


dist/info/emux_gb_libretro.info
```diff
@@ -10,3 +10,8 @@ license = "GPLv2"
 permissions = ""
 display_version = "Git"
 supports_no_game = "false"
+firmware_count = 1
+firmware0_desc = "DMG_ROM.bin (Game Boy Boot ROM)"
+firmware0_path = "DMG_ROM.bin"
+firmware0_opt = "false"
+notes = "(!) DMG_ROM.bin (md5): 32fbbd84168d3482956eb3c5051637f5"
```
RobLoach: Could we rename this in the core to match the other GB cores? `dmb_boot.bin`

Where is the source for the EmuX GB core?

esoptron: Sorry for the super late reply.

The Emux cores are all compiled from https://github.com/libretro/emux. Each core has a different build flag.

https://github.com/libretro/libretro-super/blob/master/recipes/windows/cores-windows-x64_seh-cross

I found a reference to DMG.ROM.bin here https://github.com/libretro/emux/blob/master/controllers/mapper/gb_mapper.c

RobLoach: Nice find.... Submitted PRs at https://github.com/libretro/libretro-super/pull/667 and https://github.com/libretro/emux/pull/2

PR #577, opened 2017-11-02
[Travis] Missing .ra file for Linux

RobLoach:
Getting some failures on Linux when building:
https://travis-ci.org/libretro/vice-libretro/jobs/296038702

Tries to find a .ra file, but it doesn't list.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -1251,7 +1251,11 @@ buildbot_pull(){
 
 		echo
 		echo RADIR=$RADIR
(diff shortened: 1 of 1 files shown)


orbea:
@RobLoach I silenced this error here. https://github.com/libretro/libretro-super/commit/f8e5eba8ebf211a2ae43215b6880bf784bb3dea1

That loop was only needed if a `.ra` recipe file exists so it now returns early if the file does not exist.

RobLoach:
Yeah, this isn't the solution we wanted. Thanks! :+1:

PR #588, opened 2017-11-13
libretro-buildbot-recipe.sh: Don't error if /vars does not exist.

orbea:
Silences a inconsequential error if the script tries to remove `$TMPDIR/vars` and it does not exist.


orbea:
I screwed up the commit message, it should read `$TMPDIR/vars` and not `/vars`...oh well.

PR #597, opened 2017-11-14
libretro-buildbot-recipe.sh: Combine the audio and video filter functions.

orbea:
These functions are largely identical and its easier to handle them as one.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -1031,8 +1031,8 @@ compile_filters()
 	esac
 
 	echo "compile $FILTER filters"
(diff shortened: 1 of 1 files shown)


orbea:
I added a second commit to remove the use of `make -C`. After reading the POSIX spec on `make` I found `-C` is not included... I rather not assume all of the implementations of `make` that the script supports have `-C` so it will now `cd` to the `$FILTERDIR` in a subshell to avoid the earlier `cd` maze mess. Once the subshell concludes the script will return to its previous directory.

orbea:
See here too.
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

PR #600, opened 2017-11-16
libretro-buildbot-recipe.sh: Use the $RECIPE variable

orbea:
This makes it so that we don't have to worry about `$1` being overwritten.


twinaphex:
Can you resolve the conflict?

orbea:
Its resolved now.

PR #611, opened 2017-11-22
Cleanup

orbea:
This does two things.

1. It removes `$CORES_BUILT`. This variable is only true if `$BUILD` or `$FORCE` are also true and if either of those variables are true then it will reach the same result as if `$CORES_BUILT` is true.
2. It sets `$TMPDIR/log/${BOT}/${LOGDATE}/${LOGDATE}_RetroArch_${PLATFORM}.log` as `$LOGFILE` to prevent the need to reference the whole file path many times over.


orbea:
Hold off on this PR. I need to work out some details.

orbea:
Nevermind, I rebased without the problematic commit for now. Ignore `1.` in the first comment.

Issue #610, opened 2017-11-21
Citra won't build


DukeSkinny:
Using mingw64, and running "**$ FORCE=YES SINGLE_CORE=citra ./libretro-buildbot-recipe.sh recipes/windows/cores-windows-x64_seh-noccache**" in libretro-super I get the following:

```
Setting: PATH /mingw64/bin
Setting: MSYSTEM MINGW64
Setting: platform MINGW64
Setting: PLATFORM windows
Setting: ARCH x86_64
Setting: MAKE make.exe
Setting: CMAKE cmake.exe
Setting: CC gcc
Setting: CXX g++
Setting: CXX11 g++
Setting: CORE_JOB YES
PLATFORM: win
ARCHITECTURE: x86_64
TARGET: win
SCRIPT: /home/metam/libretro-super/libretro-buildbot-recipe.sh
buildbot starting
--------------------------------------------------
Variables:
CC gcc
CXX g++
STRIP
DISTDIR /home/metam/libretro-super/dist/win_x64
JOBS 4
 
 
 
 
buildbot job started at: Tue Nov 21 19:40:18 CET 2017
 
buildbot job: recipes/windows/cores-windows-x64_seh-noccache processing citra
--------------------------------------------------
Variables:
URL https://github.com/libretro/citra.git
REPO TYPE SUBMODULE
ENABLED YES
COMMAND CMAKE
MAKEFILE Makefile
DIR libretro-citra
SUBDIR build
 
 
pulling changes from repo https://github.com/libretro/citra.git...
Already up to date.
resetting repo state https://github.com/libretro/citra.git...
HEAD is now at 4b0ec7ec Additional cleanups for clang
Removing build/
': not a valid identifierline 88: export: `dashless
buildbot job: building citra
--------------------------------------------------
CORE_JOB=YES
--------------------------------------------------
--------------------------------------------------
BUILD CMD: cmake.exe  -G "MSYS Makefiles" -DENABLE_LIBRETRO=1 -DENABLE_SDL2=0 -DENABLE_QT=0 -DDISABLE_LIBPNG=1 -DCMAKE_BUILD_TYPE="Release" --target citra_libretro
CMake Error: Could not create named generator "MSYS
 
Generators
  Visual Studio 15 2017 [arch] = Generates Visual Studio 2017 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 14 2015 [arch] = Generates Visual Studio 2015 project files.
[...]

PR #612, opened 2017-11-22
libretro-buildbot-recipe.sh: Fix passing Makefile Generators to cmake.

orbea:
Fixes https://github.com/libretro/libretro-super/issues/610

A stupid edge case...the solution is kind of ugly, but considering the recipe file structure and cmake I don't think it can be much better.


orbea:
Actually there is a better way and this was all my fault...

orbea:
Actually there is a better way and this was all my fault. Its been rebased now...

PR #613, opened 2017-11-22
libretro-buildbot-recipe.sh: Clean up

orbea:
This sets more log file names to variables like `$CORE_LOG` and `$CORE_JNI_LOG`. This helps clean up some more of these long lines.


orbea:
Actually this overlooks something...

PR #615, opened 2017-11-24
recipes: Clone the glsl shaders for RetroArch on linux.

orbea:
It was brought to my attention that the RetroArch recipe file for linux does not clone the glsl shaders and that some users needed to clone them manually. Sow now `libretro-buildbot-recipe.sh` will grab the glsl shaders too. This is only for linux because I am not sure what other platforms can use them?


fr500:
Windows, iOS, Android

hizzlekizzle:
MacOS, as well.

orbea:
Thanks, I will add glsl shaders to those after https://github.com/libretro/libretro-super/pull/616 is merged/closed to avoid merge conflicts.

PR #622, opened 2017-11-30
libretro-buildbot-recipe.sh: Check if the HEAD commit is new or not.

orbea:
Newer git versions now report `Already up to date.` instead of `Already up-to-date` so we need a more robust solution.


orbea:
Also this started happening for me after Slackware64-current updated to `git-2.15.1`.

PR #623, opened 2017-11-30
libretro-buildbot-recipe.sh: Clean cores after building them.

orbea:
If a core builds it will not reset and clean the git repo to save space on the buildbot server. If it does not build it will do nothing as before.


orbea:
Hold off on merging this until there is a consensus on this issue, thanks.

orbea:
I rebased the commit to not `git reset --hard FETCH_HEAD` and only `git clean -xdf` after builds.

I suggest that @bparker and @fr500 agree on this solution before merging. Personally it does not matter to me and I'm fine with w/e the consensus chooses.

PR #629, opened 2017-12-02
libretro-buildbot-recipe.sh: Always use the branch in the recipe file.

orbea:
If the checked out branch does not match the branch name in the recipe file the script will now make a shallow fetch of the correct branch, checkout the new branch and then remove the old one. If this is done it will also build a core with the new branch.

The motivation behind this is that I noticed that some of the `bsnes-libretro` recipes are using the incorrect and outdated branch `master` instead of the default `libretro`. While fixing this is easy, before this commit it would of also have meant changing each clone on the buildbot server manually and this now should be an automatic process. Additionally I wonder if this will uncover any repos which have been on the wrong branch for a while?

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -748,7 +748,7 @@ while read line; do
 			UPDATE="NO"
 		fi
 
-		cd "$DIR"
+		cd "$DIR" || { echo "Failed to cd to ${DIR}, skipping ${NAME}"; continue; }
 
 		CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -752,7 +752,7 @@ while read line; do
 
 		CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 
-		if [ "${GIT_BRANCH}" != "${CURRENT_BRANCH}" ]; then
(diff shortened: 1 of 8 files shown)


orbea:
I added a second commit to fix the bsnes recipes to make it easy to merge this all at once.

orbea:
@bparker and / or @fr500 Would you mind reviewing this and making sure I am not overlooking anything?

twinaphex:
OK, I'll hold off on merging until either @bparker06 and/or @fr500 have looked into this.

bparker06:
@orbea I think this might break Travis. What happens there is that when a PR needs building, it's checked out as a sort of hidden branch on the *upstream* repo, then moved into libretro-super under the same folder name tnat the recipe would have fetched, and then it builds with libretro-buildbot-recipe.sh.

orbea:
Yes, this would of probably broken that workflow. Good thing you pointed it out! I'm going to leave this here for a little bit while I think over what can be done. Someone might have to go around making sure all the repos are on the right branch manually...

orbea:
@bparker06 I rebased the PR against master and added a new commit. This commit exports `TRAVIS=1` in which I think are the right travis scripts and if this is true the buildbot will not try to change branches. Do you think this is better?

bparker06:
Yes that should work. Thanks.

orbea:
@twinaphex Since bparker seems okay with this can it be merged? It will conflict with other commits I have been working on and I will need to rebase in my local branches before I can consider pushing.

orbea:
@bparker06 I added one more commit to make this even safer. Last night a recipe failed because the repo was never cloned for some reason... In the event this occurs it can be caught early now and skipped.

PR #635, opened 2017-12-06
libretro-buildbot-recipe.sh: Remove seemingly unneeded and non-portable bash redirection.

orbea:
This should help the following issue reported by aaa801 in irc.
```
# Pastebin AyhdxPKp
Compiler: CC="ppu-lv2-gcc.exe" CXX="ppu-lv2-g++.exe"
ps3
ps3
=== 81
Building 81...
/g/retroarch/libretro-super/libretro-build-common.sh: cannot make pipe for process substitution: Function not implemented
/g/retroarch/libretro-super/libretro-build-common.sh: cannot make pipe for process substitution: Function not implemented
/g/retroarch/libretro-super/libretro-build-common.sh: line 287: >($cmd_tee -a $log_module): ambiguous redirect
```
@bparker06 I can't figure out what `exec > >$cmd_tee -a $log_module)` does that `exec > $logfile` doesn't already, but maybe you see something I am overlooking? Given the latter should be more portable I think that is preferable.

### DIFF
--- a/libretro-build-common.sh
+++ b/libretro-build-common.sh
@@ -282,12 +282,7 @@ libretro_build_core() {
 		exec 6>&1
 		echo "Building ${1}..." >> $log_module
 
-		# TODO: Possibly a shell function for tee?
-		if [[ -n "$LIBRETRO_DEVELOPER" && -n "${cmd_tee:=$(find_tool "tee")}" ]]; then
-			exec > >($cmd_tee -a $log_module)
-		else
-			exec > $log_module
-		fi
+		exec > $log_module
 	fi
 
 	case "$core_build_rule" in


orbea:
Also in case its not clear, `$cmd_tee` = `tee`.

bparker06:
I have no idea what `> >` does or even how to google that.

orbea:
Yea, I couldn't google it either... Best I found was that `exec > $logfile` is a good way to avoid the following.
```
echo > $logfile
echo >> $logfile
echo >> $logfile
```
Where instead it can do.
```
exec > $logfile
echo
echo
echo
```
It seems to create the logs correctly at least and aaa801 reported it works for him in irc. I'm still not sure what that fancy use of `tee -a` accomplishes, but it also does not seem necessary....

twinaphex:
Hi there @fr500, would you perhaps be able to answer the question?

twinaphex:
I asked radius, he didn't know either. In light of that, I'd say we just go ahead and merge it.

PR #638, opened 2017-12-07
libretro-buildbot-recipe.sh: Move build_libretro_leiradel_makefile in…

orbea:
This moves one of the last build functions into `build_libretro_generic_makefile`. I don't really have a way to test these specific recipes, but it should not break other functionality.

@leiradel Since `build_libretro_leiradel_makefile` has your name, can you review this to make sure it still works for you?


leiradel:
I'm afraid I don't have the slightest idea what's going on there... Maybe @fr500 can take a look?

orbea:
Going to close this for now, on further thought there are better ways to address this, but other steps will need to be done first.

PR #640, opened 2017-12-12
Fix Mr.Boom database name

RobLoach:
Discovered that having `Mr.Boom` as the database name ends up cropping out the `.Boom` in the menu:

mrboom

Remove the space to result in `MrBoom` fixes it.


baxysquare:
@RobLoach & @alfrix Thanks so much for fixing this.  When I made the icons, I had no idea whether to name them "MrBoom", "Mr.Boom" or "Mr. Boom" which was "a swing and a miss".   If there's a better way to know how icon names correlate with the database rather than to "name and hope for the best," let me know and I'll add it to my workflow.

RobLoach:
@baxysquare Hah! You dun good! Seems to be working now.

RobLoach:
@baxysquare Sent you a PR over at https://github.com/libretro/retroarch-assets/pull/196 .  The `.` was messing it up. It's not you, it's RA. Mr.Boom was never in the playlist before, so this is the first time we'll see if there!

PR #643, opened 2017-12-17
(Has dependancys in libretro-toolchains and RetroArch) Update 3ds toolchain path

meepingsnesroms:


meepingsnesroms:
Was already done separately.

PR #650, opened 2017-12-21
Revert "libretro-buildbot-recipe.sh: Remove seemingly unneeded and non-portable bash redirection."

orbea:
Reverts libretro/libretro-super#635

My idea had issues unfortunately, see. https://github.com/libretro/libretro-super/pull/646

I suggest reverting this for now and accepting that some older bash versions may not have this functionality. I'll try to figure out a better way, but its not nearly easy as I had hoped.

### DIFF
--- a/libretro-build-common.sh
+++ b/libretro-build-common.sh
@@ -282,7 +282,12 @@ libretro_build_core() {
 		exec 6>&1
 		echo "Building ${1}..." >> $log_module
 
-		exec > $log_module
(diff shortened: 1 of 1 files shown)


orbea:
Right now the immediate choices seem to be.

1, Revert this and live with a little less portability.
2. Live with less verbosity for now and wait for I or someone else to think of a better way. This might require some refactoring...

It doesn't really matter to me, I just don't want to make things worse...

PR #652, opened 2017-12-21
libretro-buildbot-recipe.sh: Build all three mame2014 cores at once.

orbea:
This adds three commits, the first two are minor and are included to avoid merge conflicts.

The last commit should allow to have only one `mame2014` recipe per file and it will build all three libretro cores, `mame2014`, `mess2014` and `ume2014`.

Some side effects:

1. `recipes/blackberry/cores-qnx-generic` was previously missing the `mess2014` and `ume2014` cores.
2. In `recipes/linux/cores-linux-armhf-generic` only the `ume2014` sets `PTR64=0` and `PARTIAL=1` while neither `mame2014` or `mess2014` does. I'm not sure if this is intended, but assuming it is I think fixing this in the `mame2014` Makefile would be the best solution. Alternatively I can accommodate in this script by adding a hack. The buildbot should be able to test this either way.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -323,7 +323,12 @@ build_libretro_generic_makefile() {
 	eval "set -- $CORE"
 	for core do
 		NAME="$core"
-		[ "${COMMAND}" = "BSNES" ] && ARGS="profile=${core##*_}"
+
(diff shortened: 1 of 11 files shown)


orbea:
Also, it is no longer required to build the buildtools separately so I am removing that step.

PR #654, opened 2017-12-22
recipes: Try to fix mame2014 for armhf.

orbea:
This restores the `PTR64=0` which was used for `ume2014`. `mame2014`, `mess2014` and `ume2014` now all fail due to the same error and my guess is this will help. However I will need to test it with the buildbot...

See
```
obj/libexpat.a: error adding symbols: File format not recognized
collect2: error: ld returned 1 exit status
Makefile:909: recipe for target 'mame2014' failed
make: *** [mame2014] Error 1
```
http://p.0bl.net/119116


orbea:
My guess was wrong, it still fails on the same error. I suppose this was likely broken before too.

orbea:
Seems this was broken for a long time. Last update for `mame2014` and `mess2014` on the buildbot was over a year ago. There doesn't seem to be an `ume2014` build either.

PR #656, opened 2017-12-22
libretro-buildbot-recipe.sh: Skip build if the .git directory is broken.

orbea:
The msvc2005 virtual mafchine had an issue where several of the cloned repos were corrupted and left `.git` in a non-useful state and this caused an infinite loop in the buildbot script. Now if `.git` exists and is broken it will print a warning and remove the repo instead of using it. When the buildbot script next encounters the recipe it will hopefully correctly clone the repo again.

Additionally it will now `git pull` earlier in the script and entirely avoid unintended issues when the branch in the recipe does not match the current broken branch.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -602,6 +602,12 @@ while read line; do
 			git clone --depth=1 -b "$GIT_BRANCH" "$URL" "$DIR"
 			BUILD="YES"
 			UPDATE="NO"
+		else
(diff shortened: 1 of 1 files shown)


orbea:
Also it will explicitly use the correct `.git` directory. If its broken `git` will fallback to the `.git` directory in the parent directory...

See https://stackoverflow.com/questions/5083224/git-pull-while-not-in-a-git-directory

PR #657, opened 2017-12-22
libretro-buildbot-recipe.sh: Don't rely on the UPDATE variable.

orbea:
This is a minor cleanup commit. It removes the `$UPDATE` variable and uses a more explicit use of `git` to achieve the same behavior. It should be a little safer if anything.


orbea:
I added another commit to set some variables earlier so that these variable will actually keep store the old variables.

PR #662, opened 2017-12-23
libretro-buildbot-recipe.sh: Allow building specific bsnes and mame c…

orbea:
This could be useful for travis to capture regressions if the only one of the cores is affected. Additionally it is required for fixing travis with mame2014. If the core is not bsnes-libretro, bsnes-mercury or mame2014 the $CORE variable in the travis scripts will be ignored.

When this is merged I can do the work to make sure bsnes-libretro, bsnes-mercury and mame2014 use it.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -573,7 +573,9 @@ while read line; do
 	ARGS="${ARGS# }"
 	ARGS="${ARGS%"${ARGS##*[![:space:]]}"}"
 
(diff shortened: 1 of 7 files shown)


orbea:
@bparker06 Are you okay with this?

orbea:
One caveat I have not thought about yet. This will not work if SINGLE_CORE is not set. I think I can at least prevent that from blowing up...

orbea:
I added a commit to prevent that from being an issue. This will ensure that unless `SINGLE_CORE` is set it will always use the default list of cores.

orbea:
I added an commit that will ensure that unless `SINGLE_CORE` is set it will always build the default list of cores.

PR #670, opened 2017-12-28
recipes: Try to fix emux cores for cores-linux-armhf-generic.

orbea:
These cores last built for `armhf` in 2015...the problem is that the `LEIRADEL` command requires the first `$ARGS` to be the makefile suffix where its missing here and resulted in the following issue. Additionally it uses `$PLATFORM` to help determine what makefile to use and in this case it should be `linux` not `unix`. Reading the script I don't think changing the platform here should break anything and other `.conf` files already set it as `linux`.
```
CLEANUP CMD:  make -f Makefile.unix_MACHINE=nes platform=unix_MACHINE=nes -j6 clean
make: Makefile.unix_MACHINE=nes: No such file or directory
make: *** No rule to make target 'Makefile.unix_MACHINE=nes'.  Stop.
--------------------------------------------------
BUILD CMD:  make -f Makefile.unix_MACHINE=nes platform=unix_MACHINE=nes -j6
make: Makefile.unix_MACHINE=nes: No such file or directory
make: *** No rule to make target 'Makefile.unix_MACHINE=nes'.  Stop.
COPY CMD: cp -v emux_nes_libretro.unix_MACHINE=nes.so /home/buildbot/buildbot/linux_armhf/dist/unix//MACHINE=nes/emux_nes_libretro.so
cp: cannot stat 'emux_nes_libretro.unix_MACHINE=nes.so': No such file or directory
```
@fr500 Does this look good to you?

### DIFF
--- a/recipes/linux/cores-linux-armhf-generic
+++ b/recipes/linux/cores-linux-armhf-generic
@@ -11,10 +11,10 @@ chailove libretro-chailove https://github.com/RobLoach/ChaiLove.git master YES G
 desmume libretro-desmume https://github.com/libretro/desmume.git master YES GENERIC Makefile.libretro desmume platform=armv7-neon-hardfloat
 dinothawr libretro-dinothawr https://github.com/libretro/Dinothawr.git master YES GENERIC Makefile .
 dosbox libretro-dosbox https://github.com/libretro/dosbox-libretro.git master YES GENERIC Makefile.libretro .
-emux_chip8 libretro-emux_chip8 https://github.com/libretro/emux master YES LEIRADEL Makefile libretro MACHINE=chip8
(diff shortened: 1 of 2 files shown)


orbea:
@twinaphex Appears they built. :) As I don't have the hardware someone else will have to test them.
```
14:29:22 @retrobot emux_chip8: [status: done] [recipes/linux/cores-linux-armhf-generic]
14:29:24 @retrobot emux_gb: [status: done] [recipes/linux/cores-linux-armhf-generic]
14:29:26 @retrobot emux_nes: [status: done] [recipes/linux/cores-linux-armhf-generic]
14:29:28 @retrobot emux_sms: [status: done] [recipes/linux/cores-linux-armhf-generic]
```

PR #686, opened 2018-01-07
libretro-buildbot-recipe.sh: git reset and git clean before pulling.

orbea:
Resets and cleans the repo before pulling in an attempt to help force pushing.


orbea:
Doesn't work...

PR #688, opened 2018-01-07
libretro-buildbot-recipe.sh: Make shallow fetches.

orbea:
I missed a `--depth 1` from my last PR.
https://github.com/libretro/libretro-super/pull/687

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -621,7 +621,7 @@ while read line; do
 			rm -rfv -- "$DIR" && continue; }
 
 		echo "fetching changes from repo $URL..."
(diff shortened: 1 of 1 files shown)


twinaphex:
what will this do to the git history for developers like me whom often need to bisect? 

If this will throw away all the history, you should consider making this an option instead.

orbea:
This is for the buildbot, my understanding was it should have shallow repos to preserve space and already does this in most of the script. If my assumption is wrong then this probably should not be merged, it doesn't make a big difference for me, but if you want the git history intact other changes might have to be made.

Can you discuss this with @fr500 so a consensus can be made on what the expected behavior should be?

twinaphex:
FOr the buildbot it would be OK. OK, I understand why you want this change then. Let's go ahead.

PR #705, opened 2018-01-17
ios: Remove ChaiLove

RobLoach:
Can't get it to build FreeType2 correctly. Giving up for now.


orbea:
@RobLoach You can always just disable it by changing the `YES` to a `NO`.

PR #713, opened 2018-01-24
recipes: Restore the ppsspp jni build.

orbea:
Warning: Untested.

Since cmake does not seem to be recommend for android by upstream and they use their own jni makefile I suggest use jni too.

The ideal solution might be add libretro support to the upstream jni makefile so that it stays in sync with upstream better and there are less files to maintain, but for the short term we can use the old libretro jni makefile.


orbea:
Also see https://github.com/hrydgard/ppsspp/wiki/Build-instructions

PR #728, opened 2018-02-02
port and activate remaining buildbot recipes for FreeIntv

markwkidd:
I've generally based the Makefile components for these recipes on snes9x, mame2000, and mame2003.

I believe this PR will establish recipes for every platform currently supported by the buildbot. After this we can focus on making sure these builds are working right.


twinaphex:
Do you have read-only access to our buildbot status page? If not, join us on Discord and ask bparker for read-only access.

PR #741, opened 2018-02-28
[NP2kai] remove PSP build

AZO234:
I removed NP2kai PSP build.


twinaphex:
This is not necessaey. The NO in the line already makes it skip over building this core.

AZO234:
I got it. Thanks!

PR #742, opened 2018-03-01
add LTO=true for Mesen

Tatsuya79:
Hope that works like this?


twinaphex:
@fr500 Does this work?

fr500:
It's built with MSVC LTO makes no difference there.
It only makes difference in the ming builds.

fr500:
and no, the syntax is wrong extra arguments go towards the end of the line

Tatsuya79:
ok, idk I'll just close that then.

PR #768, opened 2018-03-28
Fix and simplify android cmake path

webgeek1234:
After some more research, I found that using the toolchain file is the actually supported method of compiling android projects through cmake. I've tested compiling ppsspp against upstream plus my open PR to fix gles platforms. This is a precursor to switching ppsspp back to cmake on android.


webgeek1234:
Added a commit to strip the cmake outputs. In conjunction with https://github.com/hrydgard/ppsspp/pull/10828 this gives a usable core for Android.

PR #771, opened 2018-04-01
Android aarch64 unification

webgeek1234:
DO NOT MERGE WITHOUT TESTING

The end result of this PR is to build arm64-v8a from the main android jni recipe. It also cleans up some broken or missing cores. A side effect of one of the changes is that the main retroarch apk would build arm64-v8a support in. The last commit attempts to circumvent that while still allowing the secondary branch to build. However, I have not been able to replicate the system needed to build that apk, so this needs tested on the buildbot before merging.


webgeek1234:
Thanks to bparker for the build. It does correctly prevent arm64-v8a from being built in. This should be good to go.

PR #772, opened 2018-04-03
recipes: Build nestopia with the upstream repo

orbea:
These repos are even and we have way too many shallow forks. I think its best we let the upstream repo take the reigns so that issues and pull requests can all be submitted in one place.

Earlier work done in the buildbot script should make this change automatic. The old libretro repos will be removed and the new upstream repo will be cloned and then built.

Also bringing this to the attention of @rdanbrook.


retro-wertz:
is the use of libretro-fetch.sh (and other related scripts) are to be dropped now? i see repos moved to upstream now...

twinaphex:
I disagree with this. Leave it as it is.

orbea:
@retro-wertz I don't think this PR would of made a difference with the scripts here, but by if @twinaphex is really opposed then I guess that is that.

fwiw the main goal was to try to get all PRs and issue reports in one place and avoid juggling more than one repo, but I'm not going to push this if its going to make someone unhappy.

twinaphex:
> fwiw the main goal was to try to get all PRs and issue reports in one place and avoid juggling more than one repo, but I'm not going to push this if its going to make someone unhappy.

I think we tried this before already with libretro-meta and that turned into a huge disaster, so best we don't do that again.

Honestly, I really think we are wasting our time here on unimportant stuff, to be brutally frank. None of it will help reduce maintenance either.

orbea:
Just to be clear I meant all nestopia specific issues and PRs in one place, not all core issues in one place. I totally agree that libretro-meta wasn't the best idea.

orbea:
The only other thing I can say is that I'm fine with w/e the consensus chooses, my main interest is seeing nestopia is well maintained. :)

PR #787, opened 2018-04-14
db: Don't build SNES Hacks

RobLoach:
The Hacks database was removed. It was full of GoodSNES, which had a bunch of bloat.


RobLoach:
This is good to merge. Having a split Hacks RDB broke scanning in a few cases. There are no longer split Hacks DAT files.

bparker06:
What's the status on this? @twinaphex any input?

PR #794, opened 2018-04-25
retroarch-release.sh: Added script to create RetroArch release tarballs.

orbea:
This script will create a `.zip`, `.tar.gz` and `.tag.xz` from the most recent RetroArch release tag.

One draw back of this script is that it explicitly removes `retroarch/` and then clones it again with `libretro-fetch.sh` to ensure its always a clean and fully up to date repo. Besides being a bit slower this is not really a problem for something that should be run once per RetroArch release. It could maybe be improved once I or someone else spends time looking at `libretro-fetch.sh` more.


orbea:
On second thought `.tar.gz` is just too large to use here. With `v1.7.2` it creates a 678.05 MB `.tar.gz` and a 136.23 MB `.tar.xz`. I've updated this script accordingly.

PR #795, opened 2018-04-25
Add new Thomson TO8D core to build bot

Zlika:
Hi!

I would like to add theodore, a Thomson TO8D emulator, to libretro-super. I don't know if it is the right way to do so, please give me some guidance if I've done things wrong.
In particular I have two questions:
* this core should be able to compile and run on most platforms, however I cannot test them all by myself: can I add this core to all the recipes, or should I add it only for platforms on which I was able to test it?
* info files have some information that are redundant to the ones already included in the cores (through retro_get_system_info()). In particular, how do you keep in sync the version number?

Regards

### DIFF
--- a/dist/info/theodore_libretro.info
+++ b/dist/info/theodore_libretro.info
@@ -1,0 +1,12 @@
+display_name = "Thomson - TO8D (Theodore)"
+authors = "T. Lorblanches"
+supported_extensions = "fd|k7|rom"
+corename = "Theodore"
+manufacturer = "Thomson"
+categories = "Emulator"
+systemname = "Thomson TO8D"
+license = "GPLv3"
+permissions = ""
+display_version = "0.9"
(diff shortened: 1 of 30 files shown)


twinaphex:
Hi there, thanks very much for your contribution! We'd be glad to include it.

To answer question 1, we can just let it run for all platforms. It will likely not compile for a bunch of platforms, I can then communicate with you on strategies we can get them to compile and work. Best way to find out is to just let it try to compile the core for as many platforms as possible.

To answer question 2, keeping version number in sync is unfortunately something you'd have to do manually. These values are there because especially on systems where there is no dynamic linking available, this is the only means by which the frontend (RetroArch in this case) has any ability to tell what extensions are supported by which core. It's just a necessary evil really to provide these details.

twinaphex:
Either that or you just type in 'git' for version.

PR #796, opened 2018-04-26
theodore: change display_version to git

Zlika:
Change display_version to git in theodore_libretro.info file


Zlika:
Hi @twinaphex.
Putting "Git" in display_version only gives the hash of the commit. I would like something more human readable. For exemple in my makefile I use "git describe --always --tags" as a version number. Is it possible to do something like this in the info file?
Thanks.

PR #800, opened 2018-05-01
(MSYS2) Manually copy qwindows.dll from the Qt installation directory

rz5:
This has not been tested.


bparker06:
@rz5 The `windeployqt` command right below already copies this library (and many more) into the right place. `mingw64` should not be hardcoded as it breaks 32-bit builds, and the `qwindows.dll` file needs to go into a `platforms/` folder which the deploy command already handles. I will go ahead and revert this.

PR #808, opened 2018-05-23
Update cannonball_libretro.info

Tatsuya79:


Tatsuya79:
Well, it's a bit more complicated but works for standard play (world tracks) just with the outrun rom.

If you want to play the Japanese tracks additional roms are required.
I took them from outrundx as they seemed to have the right name but they're not recognized.
Had to find them somewhere else to run those tracks, previous versions or something?

PR #821, opened 2018-06-10
Disable bnes and meteor.

Tatsuya79:
I hope I got them all.


orbea:
@Tatsuya79 and @twinaphex Why where these disabled? They work last time I checked.

hizzlekizzle:
They work, but they're n00b-traps. That is, there's no reason to use them over any of the other options, and there are significant reasons not to use them that are known to us, obviously, but not to an outsider.

PR #826, opened 2018-06-19
disable higan_sfc_balanced again

Tatsuya79:
It's back to spamming despite the git address change.


twinaphex:
@bparker06 Try to check out what is going wrong here.

PR #831, opened 2018-06-28
disable higan_sfc_balanced

Tatsuya79:
It's compiling again and again for no reason.


twinaphex:
Ugh, shame that this is causing so many issues.

PR #845, opened 2018-07-20
windows/cmake: this needs to be set too

aliaspider:


bparker06:
It probably doesn't matter but just FYI "recipes/windows/cores-windows-x64_sjlj-generic.conf" has just `cmake` while all the others have `cmake.exe`

aliaspider:
yeah I simply followed the logic of the file I was editing. all had make.exe for MAKE except that one.

PR #848, opened 2018-07-21
libretro-buildbot-recipe.sh: Clean up.

orbea:
Should not differ in behavior, just clean up.


twinaphex:
Thanks, although there does seem to be a conflict that first needs to be resolved before we can merge this.

orbea:
Conflict resolved.

PR #860, opened 2018-08-21
general: added rpcname as an attribute for libretro/RetroArch#7089

winneon:
This PR adds `rpcname` as a new core info attribute for most available cores. This attribute will be used to discern what default platform each core emulates in an alphanumeric & legible format. The reasoning behind this attribute can be found at https://github.com/libretro/RetroArch/pull/7089#issuecomment-414502307.


RobLoach:
Rename to `systemid`? Seems like a machine name for the `systemname`.

winneon:
@RobLoach Seems like a good idea; it has been changed for a more wider use.

winneon:
@alfrix The mpv and pocketcdg icons have already been changed to movie/media before your review. As for the dolphin icons, they should stay as `gamecube` because `systemid` is supposed to reflect the platform as a whole, not just as a use case for Discord RPC icons. The same is done with Genesis Plus GX: it runs most Sega 8/16-bit platforms but it uses `mega_drive` as its `systemid` because it's the primary platform for the emulator and also the most widely used. If people get irritated that it shows a GameCube icon when they're playing a Wii game, that should be addressed in the RetroArch or icon PRs, not here.

3D Engine doesn't need a `systemid` because `systemid` defaults to `core` anyway if it's blank or not available. No need to manually set it to `core`.

PR #881, opened 2018-10-12
Add a safety check to retroarch-release.sh

gblues:
== DETAILS
After running this script and inadvertently nuking my fork, I've
added a safety check so that it doesn't happen to anyone else.

Change summary:
- update the .gitignore to ignore wiiu artifacts and Mac OSX junk files
- add two command-line parameters to `retroarch-release.sh`:

  * `--clean` will execute the legacy behavior of deleting and re-
    cloning `libretro/RetroArch`
  * `--force` will execute the release script using the current
    state of `retroarch/` without re-cloning, although it will still
    do a `git pull` and fetch submodules.

  If neither argument is provided, then the script fails with
  instructions if it detects that the `retroarch/` directory already
  exists.

@twinaphex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,6 @@
 /dist/unix/
 /dist/win*/
 /dist/osx*/
+/dist/wiiu
 /log/
+.DS_Store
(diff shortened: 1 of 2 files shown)


twinaphex:
Yeah, sorry about this.

orbea:
@gblues This script is for making tarballs for RetroArch releases, nuking your directory is intentional and you should not run this script unless you are making new tarballs for a RetroArch release.

PR #882, opened 2018-10-15
player-rules: Replace ./fetch-submodules.sh

RobLoach:
`./fetch-submodules.sh` no longer exists? Replace it with `git submodule update --init`


twinaphex:
But we don't use any git submodules in RetroArch, and have no intention to use them either.

RobLoach:
If we just remove this entry, will player-rules.d still function?

twinaphex:
We need to investigate why it's no longer there instead of just getting rid of it.

PR #884, opened 2018-10-17
Fix corename for mame2003

gblues:
== DETAILS

A recent change in MAME 2003 changed the internal corename to be
simply "mame2003". This ended up breaking netplay, because netplay
uses that internal name to map to the core filename; and with the
name no longer matching, the core lookup fails.

My first idea was to change the internal core name back, but that
plan was rejected. So this is the next safest way to fix the problem.
And really, it needs to change because the version string in the
internal name was redundant anyway.

@twinaphex @markwkidd


markwkidd:
I'd like to tag @dankcushions who has AFAIK the best handle on the consequences of these name changes for existing users. 

Things I know of that the library name effects:
  * matchmaking between static and dynamically compiled versions of RetroArch
  * the paths where game-specific overrides are saved by RA

Standardization with other cores does seem to be a good idea but I want to point out at least some of the factors involved.

fr500:
All I know is that for netplay this parameter needs to match with the internal core name.

https://github.com/libretro/libretro-super/commit/7880f177b9a8e1781bff096e0b9ad2289aa6d74b#diff-d0c047cdbfaf33cb5cc77937c894e0a5R4

fr500:
@markwkidd matchmaking doesn't care if static or dynamic linked.
But yeah as you all said, it affects overrides, remaps and shader presets.

dankcushions:
closed during hack-gate 2020. @twinaphex please re-open :)

PR #895, opened 2018-11-06
Add Mu to buildbot makefiles

meepingsnesroms:


fr500:
hey are you sure it will build and link on all those platforms?
because if it doesn't it will cause all static platform builds to fail, maybe start with linux, windows, android instead and then keep adding the static builds?

If you have tested though, all good

meepingsnesroms:
I havent tested all platforms but theres no file I/O or platform dependencys, so it should work.
If it doesnt I can then fix or disable those platforms, if anything was wrong the logs would be useful too because I dont have every SDK to test with.

twinaphex:
Let's take a chance on this I guess.

PR #906, opened 2018-11-15
add mame2003-plus rdb to database build script

markwkidd:


markwkidd:
oops.. RobLoach already made this PR

PR #914, opened 2018-11-24
remove MAME 2014 from the libretro buildbot

markwkidd:
I would like to suggest that MAME 2014 be removed from the buildbot.

I'm not suggesting that it be deleted or destroyed completely, but rather that MAME 2014 be considered a niche core that has shortcomings significant enough to not be offered for easy use to new users who might be attracted to its relatively recent date without knowing the drawbacks.

**Here is my rationale**:
  * As far as I know, anything that MAME 2014 can do, current libretro MAME can do as well or better, with as good or better performance.
  * As far as I know, MAME 2014 has the most significant unresolved issues of any MAME core such as [crashing due to missing internal paths](https://github.com/libretro/mame2014-libretro/issues/54), [crashes on major/iconic games](https://github.com/libretro/mame2014-libretro/issues/32), [crashes on more iconic games](https://github.com/libretro/mame2014-libretro/issues/61), [no sound on iconic games](https://github.com/libretro/mame2014-libretro/issues/33), [conflicting default input maps](https://github.com/libretro/mame2014-libretro/issues/37), and the list goes on
  * MAME 2014 [does not have an active maintainer, yet has fundamental issues no one is working on](https://github.com/libretro/mame2014-libretro/issues/64)
  * MAME 2014 doesn't have docs


twinaphex:
I dont really agree with this. There is still a big massive performance gulf between 2014 and 2016.

Id rather we just turn Core Updater into something where its much easier to organize available cores into categories and whatnot instead of just a plain list.

markwkidd:
This core is pretty dysfunctional in its current state and requires paths to manually be created so that it doesn't crash when launching content.

One alternative to retiring it to the buildbot would be to invest time/bounty money into it.

I haven't done very exhaustive testing about performance but I don't think there's much difference between this core and the current 2018 MAME core in performance. Are there any particular romsets which are known to work much slower in the current MAME core than 2014?

twinaphex:
Probably games like Soul Calibur and Tekken Tag are slower.

I agree that work on the core remains to be done. But still, that is not reason enough to call for its removal.

PR #918, opened 2018-12-06
add genereic armhf parameters for mame2003plus

grant2258:


markwkidd:
@grant2258 before I get started with any tweaks to the makefile, could you take a look at this buildbot arm error log? http://p.0bl.net/79406

does it relate to this change to the recipe?

grant2258:
dont know from what i see it was today that started

markwkidd:
your PR was just merged a day ago, which I think is when this problem started

grant2258:
just revert  it then was put in a while back i see there is two platforms defined it compiles fine locally for the pi

grant2258:
why are we tweaking to mame2003 makefile anyway the project are very much apart from each other.

markwkidd:
I'd like to port over the new MSVC2017 / UWP/XBox One targets in the short term.

I'm experimenting with rebasing all of the mame2003-plus non-driver/machine/cpu/soundcore improvements onto mame2003 so that we could easily use git to keep the two cores in sync for the libretro port and the core mame functions.

This may not be of much value but on the other hand it might be useful. I'm pretty far along with it, far enough that I'd like to have the two makefiles as comparable as possible in their current forms.

markwkidd:
Too much information, I know. Heh.

grant2258:
was just curious is all. Ill ask radius if he can take a look

grant2258:
to be honest i added this a while ago forgot about it was more for lakka users to get some arm optimization

markwkidd:
I think it's good to see if we can get some more eyes on this issue once other folks notice the build error. IMO it's no crisis to leave things as they are for a little while.

PR #922, opened 2019-01-02
Fix cannonball info to match database extension

parkerlreed:
The database looks for the epr-10187.88 file but the core info still referenced the older dummy CannonBall.game

https://github.com/libretro/libretro-database/blob/master/dat/Cannonball.dat#L15

### DIFF
--- a/dist/info/cannonball_libretro.info
+++ b/dist/info/cannonball_libretro.info
@@ -1,6 +1,6 @@
 display_name = "Cannonball"
 authors = "Chris White"
-supported_extensions = "88"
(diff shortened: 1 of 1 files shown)


parkerlreed:
Problem is you can't scan the content when it's set to "game"

The scanner looks at the core info and database at the same time. Database is set to find the "88" file but the core info only does "game". So in that setup it fails to scan it. 

Changing to the 88 makes it work in the scanner.

image

hizzlekizzle:
should this be changed to "game|88"? That way it scans and shows up in the file browser correctly?

PR #933, opened 2019-01-11
Pass linux arm-hardfloat-neon flags

kivutar:


Ghabry:
Is there a reason why the FORMAT_COMPILER_TARGET is "unix-armv7-hardfloat-neon" and not "unix"`as for any other Linux compiler? This is forwarded via platform=unix-armv7-hardfloat-neon to Makefiles and some Makefiles have special handling for unix via ``ifeq ($(platform), unix)`` and they need an update now.
Breaks at least the one core I'm maintaining (easyrpg).

hhromic:
I had the same question while studying the Makefile of ScummVM. The problem is that platform and platforms flags are combined together, e.g. `unix-armv7-hardfloat-neon`. For `ifeq ($(platform), unix)` to work, platform has to be *exactly* `unix`, so how do flags such as `hardfloat` or `neon` gets passed then?

Looking at the Makefiles, I have the impresion that unix should be matched instead like this `ifneq (,$(findstring unix,$(platform)))` ? so it partially matches the `unix-` part and then the other conditionals can partially match `armv` or `neon` and so on and so forth.

Hope there is some light shed on this :)

hhromic:
See for example the Makefile of the `snes9x` core:
https://github.com/libretro/snes9x/blob/30b46823b6e4b419497610f7b2f78b3b2a501964/libretro/Makefile#L49

Ghabry:
> The problem is that platform and platforms flags are combined together

Well if I understand this correctly the platform name is only used for choosing a "path" through the Makefile and I don't see where your hardfloat and neon are referenced in the snes9x and mupen64 Makefile.
Do you provide the Arm Neon specific flags via CFLAGS and CXXFLAGS? Because in that case there is no need to change the platform because the FLAGS will be already picked up.

Not fully related to this problem but I really dislike that the libretro Makefiles contain too much logic (when you add a new platform you have to update 100 Makefiles, awesome), in my opinion most of the stuff in the Makefiles should be environment variables which are set by some external build script (in libretro-super).

> Is this the correct/modern way to match unix nowadays?

No idea because there is no "common libretro Makefile" anywhere which you can use as a base for other libretro Makefiles. Any Makefile was created by copy-pasting it from some other libretro core which means they all deverged sooner or later.

Ghabry:
Based on the buildbot output this breaks a lot of cores because -fPIC is added inside a unix-check.

This affects at least:

- dosbox
- easyrpg
- melonds
- mesen
- mrboom
- sameboy
- snes9x2002, 2005 and 2005_plus
- vba_next
- vbam

hhromic:
> > The problem is that platform and platforms flags are combined together
> 
> Well if I understand this correctly the platform name is only used for choosing a "path" through the Makefile and I don't see where the hardfloat and neon are referenced in the snes9x and mupen64 Makefile.

Sorry, the example was to highlight that if platform contains `hardfloat`, `neon`, `armv` or any other *variant* to unix, then the `unix` path is completely ignored on these Makefiles that use `ifeq` because they try to match it only and exactly as `unix`. I believe the original intention is that if platform is `unix-hardfloat-neon`, then configuration for `unix` should be made plus any other settings for `hardfloat` and `neon`.

> Are the Arm Neon specific flags provided via CFLAGS and CXXFLAGS? Because in that case there is no need to change the platform because the FLAGS will be already picked up.

No, I was not refering to CFLAGS or CXXFLAGS, but as you said in your original post, to the values passed via `platform` to the Makefile.

> 
> If it is not via CFLAGS and CXXFLAGS:
> Well that's the libretro way but I really dislike that the libretro Makefiles contain too much logic (when you add a new platform you have to update 100 Makefiles), in my opinion most of the stuff in the Makefiles should be environment variables which are set by some external build script (in libretro-super).
> 

I also agree that the logic in libretro makefiles is kind of bloated/excessive and got very outdated between one another.

> To make that change fully compatible one will need to go through all Makefiles and add corresponding neon and arm handling for unix and handle this in ALL Makefiles.

Yes I agree with that. I wasn't proposing any particular change (yet), just trying to understand how Libretro and its components are designed.

> 
> > Is this the correct/modern way to match unix nowadays?
> 
> No idea because there is no "common libretro Makefile" anywhere which you can use as a base for other libretro Makefiles. Any Makefile was created by copy-pasting it from some other libretro core which means they all deverged sooner or later.

That is indeed very undesirable and sholud be addressed at some point.

> 
> I'm fine with `ifneq (,$(findstring unix,$(platform)))` instead of `ifeq ($(platform), unix)` but for all such invasive changes which can cause issues with builds, the person which pushed such a change should be responsible for fixing it everywhere or at least creating issues. Maybe it would be useful to have a README file somewhere which contains the names of all maintainers for all cores which will make it easier to notify everyone of critical changes.

Yes agreed too that the change is fundamnetal, but is not mandatory, in the end is up to every libretro component maintainer to use it or not. I was just enquiring about which is the preferred/recommended way to handle the Makefiles. Perhaps this is what is necessary, a complete skeleton project with best practices to implement Makefiles for cores.

> 
> **Maybe this should be discussed in chat (discord) with more devs**

Probably is the best idea, specially if senior devs can participate.

hhromic:
> Based on the buildbot output this breaks a lot of cores because -fPIC is added inside a unix-check.
> 
> This affects at least:
> 
> * dosbox
> * easyrpg
> * melonds
> * mesen
> * mrboom
> * sameboy
> * snes9x2002, 2005 and 2005_plus
> * vba_next
> * vbam

can you please confirm/clarify what is "this" in "this breaks a lot of cores" ?  are you refering to the original issue of `platform` not being exactly `unix` ?

Ghabry:
> are you refering to the original issue of platform not being exactly unix ?

Yes, they all check for exactly unix and conditionally add -fPIC based on this.

hhromic:
Exactly, and that also was happening recently in the scummvm core when platform is `unix-armv-neon`, in this case, the -shared and -fPIC flags were *not* being set because of the platform not being exactly `unix`.
I recently sent a PR (that was merged) to fix this by rendundantly adding these flags for the `armv` partial match conditional to not break anything else as I didn't write the original (rather complex) Makefile.

I guess the real issue here is that `unix` and `unix-whatever-else` are incorrectly assumed to be mutually exclusive by some Makefiles where in fact they are flavours of the same thing.  Definitively a standardisation approach is needed.

I don't believe that setting platform in this way is the issue, after all `libretro-super` is acting as a configure system that sets the platform and flags as needed. The makefiles should be revised.

I currently don't have time to get more involved, however I will be paying attention for any updates. Btw, how do you check buildbot's output? is that devs-only access?

Ghabry:
> I guess the real issue here is that unix and unix-whatever-else are incorrectly assumed to be mutually exclusive by some Makefiles 

Yepp and this "mutually exclusive" assumption is again the problem of "no common well-defined Makefile", I don't even know when it was decided that ``ifneq (,$(findstring unix,$(platform)))`` is now the proper way to check for Linux. Seems to be around two years ago, e.g. in mgba as part of "Makefile cleanup" (great commit description...) https://github.com/libretro/mgba/commit/cf8fe8cb67bb7a6b56f9a80cb0bdb8646a3dc872

> Btw, how do you check buildbot's output? is that devs-only access?

Unfortunately only available via the libretro Discord (channel buildbot)

PR #935, opened 2019-01-16
labeling: MAME 2014 is MAME 0.160 rather than MAME 0.159

markwkidd:
I've been looking into hooking up MAME 2014 DAT generation in order to get an accurate DAT for the exact driver versions it uses.

Once I looked into the commit history, I noticed that MAME 2014 was kept in sync with mainline all the way up through the release of MAME 0.160 on 2015-03-25: https://github.com/libretro/mame2014-libretro/commit/f7d408a9bb5d5f6deda5f2af7c6a3e051ecb0435

I will save a discussion of whether it would be more appropriate to call the core "MAME 2015" instead. In the meantime I suggest that `0.159` be changed to `0.160`


markwkidd:
(I edited the OP to include a link to that last commit from mainline: https://github.com/libretro/mame2014-libretro/commit/f7d408a9bb5d5f6deda5f2af7c6a3e051ecb0435)

PR #950, opened 2019-01-28
tvOS: Support for compiling for AppleTV using tvOS SDK

yoshisuga:


yoshisuga:
Holding off on this while i test using the existing iOS cores for tvOS

PR #975, opened 2019-03-06
Build NP2kai from libretro repo.

Tatsuya79:
merge after updating the libretro NP2kai repo from upstream


twinaphex:
Just as a reminder - we cannot update the libretro repo from upstream until all the console builds have been fixed.

PR #977, opened 2019-03-07
Up to toolchain ver and omit MIPS

AZO234:
I checked current Android NDK.
A few version number (used as path) was newer.
And MIPS toolchians are not found. Omit?


AZO234:
It was premature. I will withdraw.

PR #979, opened 2019-03-10
db: Update PCE CD, Sega CD/Saturn to use rom.name

RobLoach:
This makes Sega/Mega CD, Sega Saturn, and PC Engine CD run `rom.name` for the database matching. CRC sometimes don't apply to CD media, and the rom's name should be consistent with Redump naming in the DB.

References https://github.com/libretro/libretro-database/issues/423


RobLoach:
> Did you check this with @kivutar ?

@kivutar Made this change based on your comment here: https://github.com/libretro/libretro-database/issues/423#issuecomment-471197459

Would appreciate your testing, however.

kivutar:
This change will not work. What this change does is, during the building of the rdb, linking fields of the different .dat based on name matches instead of crc matches.

In my comment

> Let's switch to cue name matching instead of the serial?

I was proposing that the scanning algorithm could try to match on .cue name.

So I think we shouldn't merge this.

PR #982, opened 2019-03-13
db: Use rom.name for Dinothawr

RobLoach:
The `dinothawr.game` file contents could change, but the name will not.


retro-wertz:
why? this can be detected fine...

RobLoach:
And dinothawr.gamr shouldn't change much.

PR #986, opened 2019-03-14
(FBAlpha) Add ps3 target

barbudreadmon:


twinaphex:
Hmm, are you sure this will compile? The PS3 compiler doesn't support C++11 AFAIK.

barbudreadmon:
c++11 in fbalpha is only for the mips3 x64 dynarec (so it is disabled here, actually it is disabled for all archs because the dynarec doesn't work properly for most toolchains)

twinaphex:
OK, I will merge this then. If it doesn't compile we can discuss it and strive to find a solution.

PR #987, opened 2019-03-16
Enhance PPSSPP core version

julianxhokaxhiu:
This patch will allow PPSSPP core ( but I think also the other may obtain benefits ) to display the version not only based on the last commit hash, but also based on the tag version ( if used ).

As most of the cores use `git describe` command to obtain the commit hash, this will make sure on the next build from the build bot, every core will include also the tag version.

This patch should be retroactive doesn't matter the Retroarch version.

Last but not least, I also fixed the ppsspp Git url to point to the upstream. I guess this was a left over as the main recepites points to the upstream repo anyway.

---

**Before this patch:** d6d6dbb

**After this patch:** v1.8.0-3-gd6d6dbb77


fr500:
This will do exactly nothing for buildbot builds

julianxhokaxhiu:
Aren't these scripts used? Or is there another repo to be patched then?

fr500:
The buildbot doesn't use those at all.
Just libretro-buildbot-recipe.sh

julianxhokaxhiu:
Let me check the structure of that script then, I'll update this PR.

julianxhokaxhiu:
@fr500 Done. Feel free to review when you have some time. If it's fine for you please preserve also the other patches for whomever may want to build the cores themself.

julianxhokaxhiu:
@fr500 are you at least considering this patch set?

libretro-buildbot-recipe.sh
```diff
@@ -750,9 +750,9 @@ buildbot_pull(){
 					cd $DIR
 					git checkout "$BRANCH"
 				elif [ ! -z "$GIT_BRANCH" ]; then
-					git clone -b "$GIT_BRANCH" "$URL" "$DIR" --depth=1
+					git clone -b "$GIT_BRANCH" "$URL" "$DIR" --single-branch
 				else
-					git clone -b master "$URL" "$DIR" --depth=1
+					git clone -b master "$URL" "$DIR" --single-branch
```
RobLoach: The `--single-branch` change seems like a viable change. It is outside the scope of updating this to use the upstream repo, however. I'll let others decide whether we should keep the diff as is or recommend splitting it into two different PRs.

julianxhokaxhiu: Thanks, let me know :)

fr500:
I guess it's fine I don't really know about this git feature so I can't tell you if it's fine or not.

@orbea refactored the scripts a bunch so tagging him.

libretro-buildbot-recipe.sh
orbea: This will break the script....

Its already cloning a single branch, not every repo is using `master` and it will be slower without`--depth=1`.

orbea:
I think this is pointless and as is it will cause issues since its now cloning a lot more and its hardcoding the branch name...

julianxhokaxhiu:
@orbea just to be clear. my patch is not hard coding any branch, just changing the flag for cloning.

Regarding the "a lot more" is true, but at least clones only the one specified so it should be faster than not having the `--depth 1` flag. This is just to avoid misunderstandings.

orbea:
Sorry, I read it wrong this morning, you are right that it was already hardcoded to master.

That whole conditional looks weird, but that is a preexisting issue and I'll improve it later not to conflict with anything here.

Regardless the point with `--depth=1` is to save time during `git clone` and that is a lot of bandwidth used to clone the history of even a single branch and I suspect this will make travis slower too for some cores. Since I am not a big user of this script nor are they my servers I'll leave the final call for someone else, but it should at least retain `--depth=1` for travis builds.

julianxhokaxhiu:
Are the travis builds only for PRs or also for the final release in the buildbot?

orbea:
I think its only for PRs for cores (Where it has been hooked up).

See the `TRAVIS` variable in `libretro-buildbot-recipe.sh` and the related scripts in https://github.com/libretro/libretro-super/tree/master/travis.

julianxhokaxhiu:
I guess it is based on this line https://github.com/libretro/libretro-super/blob/9ed8efbf88cc692ac9a2ba3ed0946af5cb25153b/libretro-buildbot-recipe.sh#L647

Isn't it?

orbea:
No, `TRAVIS=0` means travis is not being used and that code will be skipped, that block of code is to automatically solve issues where the already cloned branch and set branch do not match. For example when the branch in the recipe file was changed and before this had to be rectified manually which was a pain.

orbea:
What could be helpful here is a way to fetch only the most recent tag with a shallow clone (`--depth=1`), but I am not sure that is possible?

julianxhokaxhiu:
Exactly. And that would make a lot of manual overhead. This is why I used `--single-branch`. Unfortunately is the best shot we have until today with what offers `git`.

I'll try to patch the script tomorrow with the `TRAVIS` logic inside then. Thanks.

orbea:
Actually we can fetch the latest tag only from a shallow clone, but it doesn't seem to help `git describe`.
````
git ls-remote --tags --sort="v:refname"  https://github.com/hrydgard/ppsspp.git | tail -n1 | sed 's/.*\///; s/\^{}//'
git clone --depth=1 -b master https://github.com/hrydgard/ppsspp.git
cd ppsspp
git fetch --depth 1 origin refs/tags/v1.8.0:refs/tags/v1.8.0
git describe --always
62928e1
```

orbea:
To help emphasize why we are using `--depth=1` currently.

```
$ git clone --depth=1 -b master https://github.com/hrydgard/ppsspp.git
$ du -h -d 1 ppsspp/ | tail -n1
63M	ppsspp/

$ git clone --single-branch -b master https://github.com/hrydgard/ppsspp.git
$ du -h -d 1 ppsspp/ | tail -n1
158M	ppsspp/
```
This adds up very quickly where there are maybe 70+ repos some with very large git histories (i.e. mame).

julianxhokaxhiu:
Indeed this PR should be seen as a unified effort in finding a solution that could preserve `--depth 1` and a better describe for tags. I'll continue investigate, and I totally get your painpoint. If anyone has a better idea, feel free to suggest.

julianxhokaxhiu:
Manage to came up with this script:
```sh
#!/bin/bash

TAG=`git ls-remote --quiet --tags --sort="v:refname" | tail -n1 | sed 's/.*\///; s/\^{}//'`
HASH=`git rev-parse @~`

git tag -a $TAG -m '' $HASH
```

Tested on the `ppsspp` repository:
```sh
$ git clone --depth=2 -b master https://github.com/hrydgard/ppsspp.git
$ du -h -d 1 ppsspp/ | tail -n1
64M	ppsspp/

$ git log
git log
commit 62928e13d8d5ee32451a393b2f10d719188d6b27 (HEAD -> master, origin/master, origin/HEAD)
Merge: ed1aa74 53e254d
Author: Unknown W. Brackets <checkins@unknownbrackets.org>
Date:   Wed Mar 27 19:14:34 2019 -0700

    Merge pull request #11942 from jeeeyul/master
    
    Fixes iOS exit freezing, Selective PTRACE syscall

commit 53e254d352986dc1093c620c58075189fd714a65 (grafted)
Author: Jeeeyul Lee <jeeeyul@gmail.com>
Date:   Wed Mar 27 11:01:20 2019 +0900

    Disables iOS exit animation to prevent freezing.
    Selective PTRACE sys call.

commit ed1aa748271abe2f7629279407e6210e9bb975c8 (grafted, tag: v1.8.0)
Author: Henrik Rydgård <hrydgard@gmail.com>
Date:   Tue Mar 26 10:39:07 2019 +0100

    Address feedback from #11936 (support memory types without CACHED as a backup)

$ git describe --always
v1.8.0-2-g62928e1
```

This way we emulate the last tag on the repo and yet we preserve the size of the cloned repository. If it's fine for you I can integrate that on the shell script easily.

orbea:
Clever, that should probably work fine, but it should also make sure the tag doesn't already exist before creating it. Also I would prefer to ensure `"$TAG"`, `$HASH` are properly quoted and to use `$(...)` instead. There are some issues with using the older form of shell expansion which are best to avoid.

julianxhokaxhiu:
Done. Feel free to review.

libretro-buildbot-recipe.sh
```diff
@@ -238,6 +238,18 @@ buildbot_log() {
 	fi
 }
 
+buildbot_tag() {
+	# If the repository has no tags, it returns an empty string
+	TAG=$(git ls-remote --quiet --tags --sort="v:refname" | tail -n1 | sed 's/.*\///; s/\^{}//')
+
+	if [ ! -z "$TAG" ]; then
```
orbea: I know other parts of the script do this, but please use `[ "${TAG}" ]` for variable is set conditionals.

orbea:
> Regarding if it doesn't exist before creating it, shallow clone excludes tags by default so if the repository is cloned normally they are present, if not we can safely suppose to continue tagging. The utility to tag is running only where a shallow clone is done.

We should not assume how the repos are cloned, its fully possible someone could remove a repo and then manually clone it again without --depth=1 for some reason... I think its better to remove any doubt that it will break now than to have it potentially cause issues down the line.

orbea:
I'm not certain, but there are two instances of `fetch depth 1` in `libretro-buildbot-recipe.sh` where it might need some changes? One where you pointed out earlier and it switches branches (Probably not used often) and the other when it fetches new changes and resets the repo to the latest commit (Seemed to avoid breaking considering all the things people did to their repos...).

Also as fr500 already pointed out, the changes in `script-modules/fetch-rules.sh` and `rules.d/core-rules.sh` are not used by the buildbot and as I don't know what those scripts are used for I am not going to review them.

libretro-buildbot-recipe.sh
julianxhokaxhiu: Done

julianxhokaxhiu:
@orbea I thought the buildbot starts always fresh on building cores. Isn't this the case? I get that updating the repos creates less bandwidth usage, although I'm used to DevOps principles where CI is done always upon clean environments. I think it's better to always start clean rather than creating branches of code that may behave weird.

If you still insist, I kindly ask to point me where this part of code is being done as I don't see it. Thank you in advance.

julianxhokaxhiu:
Thanks for the reply. I'll take a look as soon as possible as today I was extremely busy with my personal life.

Taking a quick look though I'm not sure how this impacts the whole flow. I think it's still a viable option to clean clone the whole repo. Could this be raised as a topic in your team? Or is this a strict no-go?

Sorry for double asking, but it would make scripts way easier to manage.

orbea:
> Taking a quick look though I'm not sure how this impacts the whole flow. I think it's still a viable option to clean clone the whole repo. Could this be raised as a topic in your team? Or is this a strict no-go?

@fr500 You may be better at answering this. My suspicion is that constantly re-cloning repos like mame even with a shallow clone repeatedly would be a unfortunate...

orbea:
I think basic logic would be inside `buildbot_tag`:

1. Check the remote repo for tags, if there are none do nothing.
2. Check the local repo for tags, if any exist compare with the remote repo or otherwise just add a new tag.
3. If a tag already exists and doesn't match the remote repo add the new tag or otherwise do nothing.

After this it should be fine to add the function after the fetch command too.

libretro-buildbot-recipe.sh
```diff
@@ -750,9 +763,11 @@ buildbot_pull(){
 					cd $DIR
 					git checkout "$BRANCH"
 				elif [ ! -z "$GIT_BRANCH" ]; then
-					git clone -b "$GIT_BRANCH" "$URL" "$DIR" --depth=1
+					git clone -b "$GIT_BRANCH" "$URL" "$DIR" --depth=2
+					buildbot_tag
 				else
-					git clone -b master "$URL" "$DIR" --depth=1
+					git clone -b master "$URL" "$DIR" --depth=2
+					buildbot_tag
```
orbea: Actually its been a while and I should of read this closer, the changes should not be done here. This is only for RetroArch assets, this won't help any core is pointless here.

Take a look at the linux retroarch-linux-x64.ra recipe file for instance.

https://github.com/libretro/libretro-super/blob/master/recipes/linux/retroarch-linux-x64.ra

With this change it will start trying to pointlessly add tags (I assume they don't exist?) for libretrodb, shaders, assets and etc.

julianxhokaxhiu:
Thanks for all the feedback. I'll try to process everything this weekend :)

libretro-buildbot-recipe.sh
julianxhokaxhiu: Hi, I actually did double check this and although you're right, I still think it makes sense to have it. It will not impact anything as the assets files are saved without any git information.

But if a core has no branch definition, and points to master, this may still be useful.

What do you think?

orbea: No cores are going to use the `buildbot_pull` function, only a few specific repos that will never need or use tags.

Ideally it would best to not handle RetroArch and core repos under different parts of the script, but I never got around to cleaning that up...

julianxhokaxhiu:
Related to your previous points:

1) It actually does it already, see https://github.com/libretro/libretro-super/pull/987/files#diff-3cd138b4905c4fc9efa673dc81d44bd8R245

2) If the tag already exists, it will just do nothing.
```
$ git tag -a "$TAG" -m '' "$HASH"
fatal: tag 'test' already exists
```
I don't see why a redundant check is needed

3) It will be the case if the new tag does not exists in the local ( but it exists in remote ). Basically already managed by Git

I personally don't think I need to touch the script again. Am I missing something?

orbea:
> I personally don't think I need to touch the script again. Am I missing something?

Yes, but I'm not going to repeat myself and go around in circles. Please re-read the earlier replies and ask specifically what parts you do not understand.

See these comments specifically. 

https://github.com/libretro/libretro-super/pull/987#issuecomment-478810956
https://github.com/libretro/libretro-super/pull/987#discussion_r271419430
https://github.com/libretro/libretro-super/pull/987#discussion_r272839210

libretro-buildbot-recipe.sh
julianxhokaxhiu: Ok I'll remove the line that, if you give me this assurance. All good :)

julianxhokaxhiu:
I got the point of your comments, and as said I'll remove that line. But about regarding the other two points ( 1 and 2 ), are we good?

orbea:
It still needs to be handled when fetching new changes and not just cloning or it will miss any new tags until the repo is removed for some reason and cloned again (Which may not ever happen). 

In the case of ppsspp cloning the repo over and over will be a significant waste of bandwidth and time because of the submodules which can't be made shallow clones (We don't know how deep to make the clone), the ffmpeg one in particular is huge!

libretro-buildbot-recipe.sh
julianxhokaxhiu: Done

orbea: I don't think you understood. You should not alter the `buildbot_pull` function in this PR at all. It doesn't do anything for any core.

julianxhokaxhiu:
@orbea Looking better at the script I decided to simplify a bit where the tag is being used and now it should be safe among all possible conditions. I also removed it from the `buildbot_pull` as it was used only to build RA and related assets, not cores.

I'd like to have a feedback from your side to know if it looks better now. Thank you in advance.

orbea:
Looks better, thanks. I have one more point to bring up, but I'll do that on the relevant line.

libretro-buildbot-recipe.sh
```diff
@@ -238,6 +238,18 @@ buildbot_log() {
 	fi
 }
 
+buildbot_tag() {
+	# If the repository has no tags, it returns an empty string
+	TAG=$(git ls-remote --quiet --tags --sort="v:refname" | tail -n1 | sed 's/.*\///; s/\^{}//')
+
+	if [ "${TAG}" ]; then
```
orbea: I think the only thing left is to make sure the old tag and new tag are not the same thing. As it stands it will attempt to create a new tag even if the existing tag is exactly the same.

julianxhokaxhiu: Let me fix it :) Thanks for pointing this out

julianxhokaxhiu: Done!

libretro-buildbot-recipe.sh
```diff
@@ -238,6 +238,20 @@ buildbot_log() {
 	fi
 }
 
+buildbot_tag() {
+	# If the repository has no tags, it returns an empty string
+	TAG=$(git ls-remote --quiet --tags --sort="v:refname" | tail -n1 | sed 's/.*\///; s/\^{}//')
+	# Get the current latest tag
+	OLD_TAG=$(git describe --abbrev=0 --tags)
```
orbea: You're going to want to add `--always` to `OLD_TAG`. If a repo has no tags the command will fail and print the following.
```
$ OLD_TAG=$(git describe --abbrev=0 --tags)
fatal: No names found, cannot describe anything.
$ echo $?
128
```
```
$ OLD_TAG=$(git describe --abbrev=0 --tags --always)
$ echo $?
0
```

julianxhokaxhiu: Done!

julianxhokaxhiu:
Any update on this?

ezar:
Apply??

julianxhokaxhiu:
@orbea is this still good to be merged?

ezar:
I use this patch for build ppsspp to jetson-nano! Works fantastic :)

orbea:
No idea, I'm not involved with the buildbot anymore ever since it moved to discord, I'm not sure this script is even used now?

julianxhokaxhiu:
What is then used now?

orbea:
I don't know, no one ever talked to me about it in detail.

julianxhokaxhiu:
@twinaphex would you mind helping us here please?

PR #996, opened 2019-03-22
Update wii

fr500:
Untested, may require some fixes like WiiU


twinaphex:
OK, understood. We can let the buildbot test for us.

PR #1002, opened 2019-03-26
db: Add Commodore Plus-4 and VIC-20

RobLoach:
@zach-morris found that Commodore Plus-4 and VIC-20 were missing:
https://github.com/RobLoach/libretro-thumbnails-check/issues/16


twinaphex:
Hi, thanks for submitting this.

@kivutar what do you think?

RobLoach:
This still applies well, and builds correctly. We have the .dat files in the database.

PR #1025, opened 2019-05-01
libretro-buildbot-recipe.sh: Switch repos for new urls.

orbea:
If the current remote url is different from the new url in the recipe it will now remove the old repo and clone the new url before building.

This issue was exposed by the previous PR (https://github.com/libretro/libretro-super/pull/1024) to change stella to the new upstream repo. I thought this was already implemented, but apparently not...

Also see further discussion in upstream issue https://github.com/stella-emu/stella/issues/377.

### DIFF
--- a/libretro-buildbot-recipe.sh
+++ b/libretro-buildbot-recipe.sh
@@ -612,7 +612,14 @@ while read line; do
 			{ echo "git directory broken, removing $DIR and skipping $NAME."; \
 			rm -rfv -- "$DIR" && continue; }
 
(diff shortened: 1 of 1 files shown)


twinaphex:
We dont want the old stella core replaced though. We need to give old stella another name since it is still much faster than the newer one. So there will be two Stella cores from now on.

orbea:
@twinaphex Would you mind rebranding the old core so that it doesn't conflict with the new one? I can add it after that.

This PR will be needed first though...

PR #1029, opened 2019-05-08
Add QUASI88 to buildbot

celerizer:
Compilation has been tested with Linux, Windows, Android, and Switch


hizzlekizzle:
Looks good. I'll go ahead and merge it, so just keep an eye on the buildbot channel and see if it fails anywhere. The statically linked consoles are the biggest thing to watch for, as they can cause the whole thing to not link properly, which kills the entire build.

PR #1122, opened 2019-08-30
update required BIOS in PUAE core info

rsn8887:
The core was just updated, so here's the corresponding info file update.


rsn8887:
Ready to merge, @twinaphex . Thanks.

PR #1157, opened 2019-11-07
remove old dosbox that is irrelevant now

fr500:
add dosbox-svn ce


hizzlekizzle:
this will leave a bunch of platforms that *had* a dosbox core with no dosbox core moving forward?

fr500:
Not really I think SVN should build everywhere.
The emscripten one never worked.

If any is not building I can fix that up.

hizzlekizzle:
ah, I missed a few of the recipes that already had svn. Looks like the iOS recipes are dosbox-less altogether, though.

PR #1161, opened 2019-11-11
add wii(u) beetle cores

f21red:


hizzlekizzle:
Since those are statically linked platforms, if they don't build, it will cause the whole set to fail to publish. I'll see if I can get anyone to test building them and if they work, I'll merge it.

f21red:
Oh~. That'd be terribly bad. xo

Cee123:
I've tried with the latest nightly on the Wii. I can confirm that it does work! I had a suspicion it would work. The only issue really is that you get a crash dump if you try to load multiple games one after the other. But other than that, it works!

PR #1168, opened 2019-11-15
Wii: add recipe for beetle lynx

tryal-star:


tryal-star:
Beetle Lynx builds okay with Wii toolchain.

PR #1172, opened 2019-11-19
fix md5sum for zip file

AlexxandreFS:
Compressed files have different md5sum depending on the program used and the compression ratio. In these files md5sum cannot be used.

Bios naomi.zip suggests hashing a file with several files, but only 4 of them are used in the emulator, see:
https://github.com/libretro/flycast/blob/64cd7e1d0a372ce904eef2b21ae793c241fc1477/core/hw/naomi/naomi_roms.h#L132


sabotendarsama:
Finally someone sent a PR to fix it.

Issue #1149, opened 2019-10-20
info: display_name: Famicom -> Super Famicom (SFC)


slacknk:
NES ~= Famicom. SNES != Famicom. For all snes-info: bsnes, snes9x &etc.
https://github.com/libretro/libretro-super/blob/master/dist/info/
```
display_name = "Nintendo - SNES / Famicom (_CORENAME_)"
```
example, [snes9x2005_libretro.info](https://github.com/libretro/libretro-super/blob/master/dist/info/snes9x2005_libretro.info): 
* `display_name = "Nintendo - SNES / Famicom (Snes9x 2005)"`

I'm about `display_name = "Nintendo - SNES / Super Famicom (_CORENAME_)"`
> https://en.wikipedia.org/wiki/Super_Nintendo_Entertainment_System
>>  In Japan, the system is called the Super Famicom (SFC).

But this name longest, maybe

```
- display_name = "Nintendo - SNES / Super Famicom (_CORENAME_)"
+ display_name = "Nintendo - SNES / SFC (_CORENAME_)"
```

hizzlekizzle:
I'm not opposed to this, but I am concerned that it will push the corename so far out that many will be lost to scrolling.

slacknk:
Ok, too https://docs.libretro.com/library/compatibility/snes/
Libretro Docs - For Users - Core Library: Emulation - Nintendo Emulation
```
Nintendo - SNES / Famicom (Beetle bsnes)
Nintendo - SNES / Famicom (bsnes Accuracy)
Nintendo - SNES / Famicom (bsnes Balanced)
Nintendo - SNES / Famicom (bsnes C++98 (v085))
Nintendo - SNES / Famicom (bsnes Performance)
Nintendo - SNES / Famicom (bsnes-mercury Accuracy)
Nintendo - SNES / Famicom (bsnes-mercury Balanced)
Nintendo - SNES / Famicom (bsnes-mercury Performance)
Nintendo - SNES / Famicom (higan Accuracy)
Nintendo - SNES / Famicom (nSide Balanced)
Nintendo - SNES / Famicom (Snes9x 2002)
Nintendo - SNES / Famicom (Snes9x 2005 Plus)
Nintendo - SNES / Famicom (Snes9x 2005)
Nintendo - SNES / Famicom (Snes9x 2010)
Nintendo - SNES / Famicom (Snes9x)
```
* SNES as Super Famicom/SFC - as the truth as well as
```
[...]

PR #1182, opened 2019-12-09
Add more info to info files pt 1

hizzlekizzle:
This organizes the core info files a bit more and adds a lot more information to them with respect to libretro features. My hope is that this info could be useful for frontends in the future.

While I was at it, I corrected where those cores are listed as "SNES / Famicom" instead of "SNES / SFC", though there are still a few more to correct. (see https://github.com/libretro/libretro-super/issues/1149#issuecomment-563089694)


twinaphex:
Seems like an interesting idea.

@hizzlekizzle Try talking some more about this with @jdgleaver.

PR #1188, opened 2019-12-20
Update info file for PUAE core

rsn8887:
This update is needed because of the recently added CD32 support in the PUAE core.


rsn8887:
Ready to merge @twinaphex

PR #1212, opened 2020-01-19
Add firmware files to Mesen-S info

Sanaki:
I copied these entries directly from the bsnes info file with Sour's confirmation (on discord). Non-split versions are supported as well, but I don't have those available and cannot confirm correct naming.


Sanaki:
Per Sour, the files in question currently are expected in the Firmware subdirectory, but that will be changed to the bare system directory to match bsnes soon (presumably in the next build). This PR reflects that change.

PR #1218, opened 2020-01-22
PCE: Disable building msvc2017 for now

negativeExponent:


hizzlekizzle:
Disabling this because it's building *all* of the cores for some reason:
http://paste.libretro.com/187209

PR #1280, opened 2020-03-14
Reenable prboom on 3ds

phcoder:
It runs fine at 60 fps


jdgleaver:
It also crashes when you exit RetroArch :)

This normally only happens if you save or load a game, though.

The prefetch abort happens at the very end of  `rarch_main()` i.e. after `return 0;` is called.

It's some very weird corruption - no one was able to work out why it happens.

I don't think we can add it to buildbot until this issue is fixed...

phcoder:
> It also crashes when you exit RetroArch :)
> 
> This normally only happens if you save or load a game, though.
> 
> The prefetch abort happens at the very end of `rarch_main()` i.e. after `return 0;` is called.
> 
> It's some very weird corruption - no one was able to work out why it happens.
> 
> I don't think we can add it to buildbot until this issue is fixed...

https://github.com/libretro/libretro-prboom/pull/133

jdgleaver:
@phcoder Fantastic! Awesome work :)

PR #1306, opened 2020-03-22
(TIC-80) Target the upstream repository

RobLoach:
Have the libretro-super scripts build using the upstream repo rather than our fork. The core is pretty much stable at this point, so we can use the upstream. Also updates the metadata with some of the new standard fields.


hizzlekizzle:
eyyy, thanks for hitting the info. I need to get back on that...

PR #1322, opened 2020-03-27
saturn: Add .bin to supported file extensions

RobLoach:
This adds `.bin` files to the allowed file extensions for all Saturn cores.

- Kronos
- ~~Mednafen~~
- Yabause
- YabaSanshiro

cc @pkos ... Found this when testing out the new scanner.

### DIFF
--- a/dist/info/mednafen_saturn_libretro.info
+++ b/dist/info/mednafen_saturn_libretro.info
@@ -1,6 +1,6 @@
 display_name = "Sega - Saturn (Beetle Saturn)"
 authors = "Mednafen Team"
-supported_extensions = "bin|ccd|chd|cue|toc|m3u"
+supported_extensions = "ccd|chd|cue|toc|m3u"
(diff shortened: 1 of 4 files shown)


hizzlekizzle:
beetle/mednafen-saturn shouldn't support bin, though, should it? That is, it can only load cue sheets, right?

dist/info/mednafen_saturn_libretro.info
```diff
@@ -1,6 +1,6 @@
 display_name = "Sega - Saturn (Beetle Saturn)"
 authors = "Mednafen Team"
-supported_extensions = "cue|toc|m3u|ccd|chd"
+supported_extensions = "bin|ccd|chd|cue|toc|m3u"
```
RobLoach:
``` suggestion
supported_extensions = "ccd|chd|cue|toc|m3u"
```

PR #1336, opened 2020-04-05
Add hatari core to the Vita recipes

FmT0:


r-type:
@hizzlekizzle Hi, any chances you can look at  this one ?  thanks in advance .

PR #1339, opened 2020-04-06
dosbox_core: Add BASS/BASSMIDI libs to core info file

realnc:


hizzlekizzle:
those will always be missing, though, won't they? that is, no one will ever have them in the system dir?

realnc:
> those will always be missing, though, won't they? that is, no one will ever have them in the system dir?

Yeah, they can't be bundled with the core due to licensing. So users will need to download them on their own.  I documented both here:

https://github.com/libretro/dosbox-core/blob/libretro/README.md

as well as in the core option description. In any event, this is all done through `dlopen()` at runtime to stay GPL compliant and thus it's purely optional.

hizzlekizzle:
In that case, would it make sense to just include them as a note instead of a present/missing firmware?

realnc:
No, because RA can't check that. Listing these files as firmware allows me to tell people to check the core info dialog instead of going through a lengthy process of directory names, upper/lower case letter verification etc for people who report that BASSMIDI isn't working for them.

hizzlekizzle:
nvm, looks like it does track them. :o

merging

PR #1341, opened 2020-04-08
Add Reicast to cores-linux-x64-generic

DimitrisVlachos:
Hello everyone,

We have now an official, but still primitive, libretro port.

I'm opening this PR to get feedback if this is the right way to do it - I think there's some CMAKE changes we also have to do (eg, add the beta target?)

How does this look, overall?

Thanks :)


hizzlekizzle:
hey, sorry for leaving this hanging. As I mentioned to skmp on discord, we were hoping to get written permission to use the reicast name, etc. for the purposes of hosting and distributing the core. I think he's working on it now.

DimitrisVlachos:
No worries for the delay. Thanks for the update. 
SKMP will prepare a contract for you to review later today.

If everything goes through , we will also put a backlink to  your website from our download page
as an official distribution channel.

Does it sound fair ?
Thanks :)

hizzlekizzle:
Yeah, sounds good to me. I'll check with some other folks but I don't foresee any issues.

PR #1348, opened 2020-04-11
Bump Gearsystem version to 3.0.0

drhelius:


drhelius:
Removing because I found a bug and I will be releasing 3.0.1

PR #1358, opened 2020-04-19
Fixed various firmware issues in info files

Sanaki:


Sanaki:
Mu also shows a full file replacement because I fixed the crlf line endings while I was at it. It was the only info file using those.

hizzlekizzle:
Looks good, thanks!

PR #1388, opened 2020-05-02
dosbox_core: Add to Linux armhf recipe

realnc:
It will download the build from github, no actual build is performed.

### DIFF
--- a/recipes/linux/cores-linux-armhf-generic
+++ b/recipes/linux/cores-linux-armhf-generic
@@ -16,6 +16,7 @@ crocods libretro-crocods https://github.com/libretro/libretro-crocods.git master
 desmume libretro-desmume https://github.com/libretro/desmume.git master YES GENERIC Makefile.libretro desmume/src/frontend/libretro
 desmume2015 libretro-desmume2015 https://github.com/libretro/desmume2015.git master YES GENERIC Makefile.libretro desmume
 dinothawr libretro-dinothawr https://github.com/libretro/Dinothawr.git master YES GENERIC Makefile .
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
I know bparker was always really skeeved out about scripts downloading binaries from external sources because it can be used as an attack vector on the entire ecosystem just by compromising a single point of failure.

Is that something we should be concerned about here, as well?

realnc:
It's an automated build done on GitHub's CI service, which runs on Azure. It then gets uploaded from the CI directly to https://github.com/realnc/dosbox-core/releases/tag/latest_build, which is where the buildbot will fetch it from.

I guess if Azure and GitHub get compromised in a way that alters binaries produced on Azure VMs, we're all screwed anyway :P

hizzlekizzle:
heh, well, my thought was if they just compromised your repo and uploaded whatever and tagged it as the latest release.

realnc:
Yes, if they compromise my account, they can do that. Are you suggesting that we remove these binary downloads from the buildbot? This is currently done for the two windows builds (x86 and x64) and the macos build.

hizzlekizzle:
hmm, well, if the others are already like that, I'll go ahead and merge this one, too (esp since I probably merged the last ones), but I think we should probably discuss with the people who do the server administration whether this is how we want to handle it long-term.

Disclaimer: idgaf either way. I just know it's the same consideration that keeps us from having appimage builds already, so I figured it prudent to bring it up.

realnc:
Even though I do have 2FA enabled on github, it's a valid concern. I think I can improve things by having the CI push the sha256 sums of the zip files into a branch that only contains these sums. When the buildbot downloads a build, the download make target can then check if the downloaded files match.

Let me see what I can come up with.

PR #1427, opened 2020-05-27
database: Add a headered directory

RobLoach:
This could contain DATs with headers in them. This affects NES, Atari 7800, etc.


twinaphex:
Has this been tested?

RobLoach:
Yes, it builds correctly.

PR #1429, opened 2020-05-30
Arcade cores info updates.

bslenul:
* Added a "description" field to the arcade cores info files so you can easily check which ROMsets are used by the cores.
* Renamed 'MAME' to 'MAME - Current', like it was done for Snes9x core. It will clear any confusion when talking about latest MAME.
* Added a really brief description for HBMAME (using the first sentence from official website, hope that's OK?), MAME 2003-Plus and FBN (just to mention it was previously FBA).

To be clear the goal here is just to have an easy way to check ROM sets version requirements thanks to the new core description addition, my english is not good enough to make a proper description. But adding something like "MAME 2003 is faster but less compatible/accurate" and other informations like this later could also be useful IMO.


hizzlekizzle:
This is a good start, thanks!

PR #1446, opened 2020-07-05
Updated Beetle Handy core name

Sanaki:
Updated core name to match the update from libretro/beetle-lynx-libretro#37. Should only be merged if that PR is merged.


hizzlekizzle:
I merged the other two, but this one is in dispute, as the standard pattern for mednafen-based cores is beetle-system, so it may not make sense to buck that trend here with beetle-origemu.

Sanaki:
My sole goal is consistency. If we're going to stick with Beetle Lynx that's fine too, though the info file should be updated to reflect that if so. Having two different names for one core just serves to confuse users.

Sanaki:
Closing since the consensus seem to be that "Beetle Lynx" is preferable. I'll submit a new PR to unify under that name.

PR #1448, opened 2020-07-05
Added hiscore.dat to FBNeo firmware list

Sanaki:


hizzlekizzle:
@barbudreadmon any thoughts?

Sanaki:
To be clear, this is in the FBNeo core repo and provides data to enable high score storage. The file itself isn't written in the process. It's a bios in the same sense as the BlastEm and Nestopia databases, which are also listed as firmware in their respective info files.

barbudreadmon:
i don't mind

hizzlekizzle:
k, thanks for the input. I'll go ahead and merge it, then :)

PR #1464, opened 2020-09-19
Add handy to Wii / WiiU cores

trioan:


trioan:
Tested and working, as noted here.
https://github.com/libretro/libretro-handy/issues/75

PR #1465, opened 2020-09-19
Remove VeMUlator core (Developer's request)

MJaoune:
I am the main (And only) developer of the SEGA VMU's core named VeMUlator, and I would like the core to be removed from RetroArch for some reasons. The core might also have some game-breaking timer bugs and sound emulation bugs. It also has no practical way to access saves (Memory cards) for other cores such as Dreamcast ones (Which is almost the main reason why people would want to use this core in the first place). It is also not tested enough and might have some RetroArch crashing glitches.

I also have other reasons why I would want it removed.

Please accept,
Thanks in advance.


twinaphex:
I'm sorry but after careful consideration we cannot honor this request, because it would inconvenience users for no real reason. In this case we have to side with our users and with what the license of this work allows for.

The emulator in specific was licensed as GPL before being relicensed as BSD. 

https://github.com/MJaoune/vemulator-libretro/blob/master/LICENSE

Permissions as stated here are:

Permissions
* Commercial use
* Modification
* Distribution
* Private use

In fact, the repo on our organization is still of the GPL version -

https://github.com/libretro/vemulator-libretro/blob/master/COPYING

We follow the license and we have not done anything wrong. You can retroactively change the license but it will have no bearing on the work as it currently exists, and all it would result in is that there is simply a hard fork. Ultimately, the work as it exists already is able to be used by anybody without any license restrictions.

PR #1466, opened 2020-09-20
add bluemsx-fmsx Wii

metallic77:
add bluemsx, fmsx as they can compile on devkitpro and run on my Wii without any modifications on source code.


metallic77 (commented):
Bluemsx and fmsx compile directly with devkitpro and are usable on real hardware.

PR #1482, opened 2020-10-31
Update platform name for Nintendo Switch

MetalNeo:
Most libretro cores do not identify the Switch's platform as "switch." Instead, most use "libnx" to identify the Switch's platform when building.

### DIFF
--- a/libretro-build-switch.sh
+++ b/libretro-build-switch.sh
@@ -14,4 +14,4 @@ else
 	fi
 fi
 
-platform=switch ${BASE_DIR}/libretro-build.sh $@
+platform=libnx ${BASE_DIR}/libretro-build.sh $@


m4xw:
Thats libtransistor :)

m4xw:
It can be argued to just remove that script tho.

MetalNeo:
Why shouldn't the default platform be changed?

Cores like [PCSX-ReARMed](https://github.com/libretro/pcsx_rearmed) or [Flycast](https://github.com/libretro/flycast) do not have options for libtransistor, only libnx. Meanwhile, most of the cores that support libtransistor also seem to support libnx, such as [Genesis Plus GX](https://github.com/libretro/Genesis-Plus-GX) or [Nestopia](https://github.com/libretro/nestopia).

Plus, the [Docs](https://docs.libretro.com/development/retroarch/compilation/switch-libnx/) also recommend using libnx as the platform.

m4xw:
There is no "default platform", this is just a dead script, nothing here uses it (and you shouldn't either)

PR #1499, opened 2021-01-26
(PUAE) CHD support

sonninnos:
Valid extensions sync with the core.


Sanaki (approved):
Simple update, looks good to go.

PR #1530, opened 2021-04-11
Update fuse_libretro.info

tomconte:
- Bump version to 1.6.0
- Allow .dsk extension
- Fix ROM file names for Scorpion


Sanaki:
The 1.6.0 update was reverted. This shouldn't be merged until the update is fixed and merged again. See https://github.com/libretro/fuse-libretro/pull/89#issuecomment-815673861 for info on the revert.

tomconte:
The revert was reverted (...) and thus the original 1.6.0 update applied, via https://github.com/libretro/fuse-libretro/pull/93

Sanaki:
So it was. My mistake.

PR #1541, opened 2021-05-02
Fuse: fix BIOS paths

tomconte:
Fix the BIOS paths so that they match what is described in the documentation. See https://github.com/libretro/fuse-libretro/issues/98


dist/info/fuse_libretro.info
```diff
@@ -20,28 +20,28 @@ database = "Sinclair - ZX Spectrum +3|Sinclair - ZX Spectrum"
...
 firmware0_opt = "true"
 firmware1_desc = "128p-1.rom (Pentagon machines)"
-firmware1_path = "128p-1.rom"
+firmware1_path = "fuse/128p-1.rom"
 firmware1_opt = "true"
 firmware2_desc = "trdos.rom (Pentagon machines)"
-firmware2_path = "trdos.rom"
+firmware2_path = "fuse/trdos.rom"
 firmware2_opt = "true"
 firmware3_desc = "gluck.rom (Pentagon 128K / 1024)"
```
zzedmore, line 31: @tomconte Just a minor naming quibble 😄 . Should be -
`firmware3_desc = "gluck.rom (Pentagon 512K / 1024K)"`
as gluck.rom isn't used on Pentagon 128K.

PR #1547, opened 2021-05-26
Update duckstation_libretro.info

stenzek:
Updates supported image formats, license, and description.

A range of BIOSes are supported (all revisions, PS2 and PS3 BIOS), but I'm not sure it makes sense to include them in here.

### DIFF
--- a/dist/info/duckstation_libretro.info
+++ b/dist/info/duckstation_libretro.info
@@ -1,10 +1,10 @@
 # Software Information
 display_name = "Sony - PlayStation (DuckStation)"
 authors = "stenzek"
-supported_extensions = "cue|bin|img|chd|pbp|m3u"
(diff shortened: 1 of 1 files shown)


dist/info/duckstation_libretro.info
```diff
@@ -1,10 +1,10 @@
 # Software Information
 display_name = "Sony - PlayStation (DuckStation)"
 authors = "stenzek"
-supported_extensions = "cue|bin|img|chd|pbp|m3u"
+supported_extensions = "exe|psexe|cue|bin|img|iso|chd|pbp|ecm|mds|psf|m3u"
 corename = "DuckStation"
 categories = "Emulator"
-license = "GPLv3"
+license = "Non-commercial"
```
dev-0x7C6, line 7: @stenzek Hi, Could you clarify license? Something changed, it's under multiple license or no longer GPL-3.0?

Regards,
Bartłomiej

PR #1553, opened 2021-06-11
moonlight: add core info for switch running lakka

plaidman:


dist/info/moonlight_libretro.info
```diff
@@ -0,0 +1,23 @@
...
+display_version = "1.0.3"
+
+# Hardware Information - Information about the hardware the core supports (when applicable)
+# manufacturer = "nVidia"
+# systemname = "Moonlight"
+
+# Libretro Features - The libretro API features the core supports. Useful for sorting cores
+# savestate = "false"
+# cheats = "false"
+# supports_no_game = "true"
```
RobLoach, line 18:
``` suggestion
supports_no_game = "true"
```

dist/info/moonlight_libretro.info
```diff
@@ -0,0 +1,23 @@
...
+# hw_render = "true"
+# is_experimental = "true"
+
+# Descriptive text, useful for tooltips, etc.
+# description = "A port of the nVidia streaming client Moonlight. Currently only tested using Lakka running on Nintendo Switch."
```
RobLoach, line 23:
``` suggestion
# Hardware Information - Information about the hardware the core supports (when applicable)
manufacturer = "nVidia"
systemname = "Moonlight"

# Libretro Features - The libretro API features the core supports. Useful for sorting cores
savestate = "false"
cheats = "false"
supports_no_game = "true"
hw_render = "true"
is_experimental = "true"

# Descriptive text, useful for tooltips, etc.      
description = "A port of the nVidia streaming client Moonlight. Currently only tested using Lakka running on Nintendo Switch."
```

plaidman: good call. updated.

dist/info/moonlight_libretro.info
```diff
@@ -0,0 +1,23 @@
...
+# Software Information - Information about the core software itself
+display_name = "Moonlight"
+categories = "Streaming client"
+authors = "Rock88"
+corename = "Moonlight"
+license = "GPLv3"
+display_version = "1.0.3"
+
+# Hardware Information - Information about the hardware the core supports (when applicable)
+manufacturer = "nVidia"
```
RobLoach, line 12: It's branded as NVIDIA in most places on their website :shrug: 

``` suggestion
manufacturer = "NVIDIA"
```

plaidman: oh yea missed that from your first comment. updated again.

dist/info/moonlight_libretro.info
```diff
@@ -0,0 +1,23 @@
+## All data is optional, but helps improve user experience.
+
```
RobLoach, line 2: Don't think we need this header here :wink: 
``` suggestion
```

RobLoach (approved):
I haven't tested the Moonlight core, so I'm not sure if it's in a workable state, but doesn't hurt to have the .info file there. Might want to check with @Rock88 to see how he feels about pushing this forwards. Thanks!

dist/info/moonlight_libretro.info
plaidman: updated.

plaidman:
We have reached out to rock88 a few times but so far he's been unresponsive. I think he's working on a proper switch port now, and may have abandoned the lakka port.

plaidman:
Though we have tested it in its current state and it works great!

PR #1560, opened 2021-07-05
Enforce gcc-10 for citra, dolphin and melonds cores in linux-x64-generic

Shoegzer:
Forcibly downgrade the compiler to gcc-10 for citra, dolphin and melonds cores as they do not currently compile with gcc-11 (although their upstream counterparts do).

Although rebasing the cores to upstream would of course be better, this PR works around the following issues:

* https://github.com/libretro/citra/issues/88
* https://github.com/libretro/dolphin/issues/215
* https://github.com/libretro/melonDS/issues/110

Note that this PR only covers linux-x64-generic.  Other recipes probably also need to be updated.


Shoegzer:
@twinaphex can you approve this soon?  This will need to be fixed in the cores themselves at some point but at least this gets them compiling for systems with gcc-11+ for now.

abouvier:
fbneo (only when USE_X64_DRC=1) needs gcc-10 too I think.

PR #1562, opened 2021-08-10
play: update supported extensions

gouchi:
[Source](https://github.com/jpd002/Play-/blob/master/Source/ui_libretro/main_libretro.cpp#L177)


Sanaki:
While I didn't test it personally, `mds` is showing as supported now as well.

PR #1563, opened 2021-08-11
Add chd to supported extensions for Play!

Sanaki:
Currently only DVDs are supported, not CDs, but I did personally verify they work fine in the core.


gouchi:
Duplicate of #1562 ;)

Sanaki:
Bugger, so it is. Missed that.

PR #1564, opened 2021-08-14
remove duplicate keys from core info files

ReenigneArcher:
There are duplicate keys in several of the core info files. I have removed the duplicates.

### DIFF
--- a/dist/info/mednafen_psx_libretro.info
+++ b/dist/info/mednafen_psx_libretro.info
@@ -27,7 +27,6 @@ memory_descriptors = "true"
 libretro_saves = "true"
 core_options = "true"
 load_subsystem = "false"
-hw_render = "false"
 
 # Firmware / BIOS
 firmware_count = 3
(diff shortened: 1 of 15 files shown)


hizzlekizzle:
good catch on these, but it's probably a good idea to remove the VICE descriptions that contain less information and keep the ones with more.

ReenigneArcher:
> 
> 
> good catch on these, but it's probably a good idea to remove the VICE descriptions that contain less information and keep the ones with more.

Probably a good idea, I agree. I updated the PR with those changes.

Also have some questions.
- I noticed there are several info files that reference databases that do not exist. Should these be removed from the info files? If so I'll make a separate PR for it.
- `systemname` is fairly inconsistent for cores that support multiple systems. Can this be made to use the actual system names the core supports, separated by a `|` like the other items are generally separated?
- `systemid`... basically same question as for `systemname`. I've noticed only one systemid is ever used even when the core supports multiple.

Probably these are not really important for most people, but reason I noticed these is I'm developing a game launcher front end and some scripts that match the cores to the platforms. `database` is the only one that is consistent, but it doesn't always have that property.

ReenigneArcher:
Anything else needed in order to merge this?

PR #1586, opened 2021-12-04
Add missing core info files

ReenigneArcher:
Added missing core info files:
- fbalpha_libretro.info

I took my best guess at this, but an expert on this cores should probably review.

I have allow edits by maintainers enabled if you'd like to make any edits yourself before merging, or let me know what I should change and I'll be happy to make the updates.


dist/info/fbalpha_libretro.info
```diff
@@ -0,0 +1,103 @@
+# Software Information
+display_name = "Arcade (FinalBurn Alpha)"
+authors = "FBAlpha Team"
+supported_extensions = "zip|7z|ccd|sub|img|iso|MODE1|2352|bin|cue"
+corename = "FinalBurn Alpha"
```
RobLoach, line 5: I believe this was renamed to FBNeo. https://github.com/libretro/libretro-super/blob/master/dist/info/fbneo_libretro.info

ReenigneArcher: The core still exists on the build bot

ReenigneArcher: Just went to double check (in case I was using old information) and it's still there, at least for Windows... I didn't check other platforms.

https://buildbot.libretro.com/stable/1.9.14/windows/x86_64/RetroArch_cores.7z

image

RobLoach: While the old cores are there for backward compatibility, you'll likely want to use FBNeo. Here's a few articles on where it's improved: https://www.libretro.com/index.php/category/fbneo/

ReenigneArcher: Thanks for your comments... I'm familiar with the history of why it was moved to a different project (https://www.google.com/search?q=capcom+home+arcade+illegal&oq=capcom+home+arcade+illegal). I still believe an info file should exist for each core that's available though. The info file is often the best way to find how to use a core properly.

RobLoach: Seems reasonable. Thanks! :+1:

RobLoach (approved):
While fbalpha is now fbneo, I think having the .info file there seems fine. Thanks!

dist/info/fbalpha_libretro.info
ReenigneArcher: > Seems reasonable. Thanks! 👍

Thank you! This will help me a lot with my front end as I'm parsing the info files to help simplify things for the end user.

PR #1587, opened 2021-12-04
Fix databases

ReenigneArcher:
- Removed databases with no corresponding rdb file
- Fixed names of incorrectly mapped database files
- Changed systemname "DS" to "Nintendo DS" for desmume2015_libretro and desmume_libretro


RobLoach (approved):
Thanks!

RobLoach (changes requested):
Looks like a conflict

RobLoach (approved):
This is great!

PR #1595, opened 2021-12-22
Add missing supported extensions for SwanStation

DarthMew:


dist/info/swanstation_libretro.info
```diff
@@ -1,7 +1,7 @@
 # Software Information
 display_name = "Sony - PlayStation (SwanStation)"
 authors = "stenzek"
-supported_extensions = "cue|bin|img|chd|pbp|m3u"
+supported_extensions = "exe|psexe|cue|bin|img|iso|chd|pbp|ecm|mds|psf|m3u"
```
RobLoach, line 4: `exe`?

DarthMew: Yes, it's actually a valid extension for PS1 executables (like for example [this](https://github.com/libretro/libretro-content/blob/master/Sony%20-%20PlayStation/PSXTEST.EXE)).

PR #1597, opened 2021-12-27
info: Fix pcsx_rearmed_neon name of .info file

### LINKED ISSUE #1163: Wrong core name `pcsx_rearmed_libretro_neon`
I'm doing a parser of the Libretro INFO files into JSON to create a set of databases for [Advanced Emulator Launcher](https://github.com/Wintermute0110/plugin.program.AEL), a Kodi emulator frontend. You can check example of the output [here](https://github.com/Wintermute0110/plugin.program.AEL/tree/release-0.9.8/dev-core/data), for example, a list of BIOSes and the cores they use the BIOS, etc.

I have discovered that the core `pcsx_rearmed_libretro_neon` has an incorrect name. For all the cores, the name can be extracted by removing `_libretro.info` with the exception of `pcsx_rearmed_libretro_neon`. I think the correct name for that core must be `pcsx_rearmed_neon_libretro`.

I would like to open a pull request to rename that INFO file but I guess some other files must be changed right?

RobLoach:
This fixes the name of the pcsx_rearmed_neon_libretro.info file.

Fixes #1163


hizzlekizzle:
oh dang, nice catch.

PR #1605, opened 2022-01-06
Remove TempGBA

phcoder:
It's included only for psp and even there it's disabled
Even on psp better alternatives are available.


RobLoach (approved):
Archived, and not a thing anymore.

PR #1606, opened 2022-01-07
Remove recipes that have been replaced by CI/CD

phcoder:


RobLoach:
While I agree we depend on GitLab recipes instead now, but this will break build systems that use libretro-super to build the cores. Unsure about their removal :thinking:

hizzlekizzle:
I think the intent was to leave these around for people who want/need to self-compile, but it is inevitable that they will get out of sync and degrade. Dunno what the best solution is here.

phcoder:
> I think the intent was to leave these around for people who want/need to self-compile, but it is inevitable that they will get out of sync and degrade. Dunno what the best solution is here.

Scripts for self-compilation don't use recipes. Instead the list of cores is in the scripts

fpscan:
Is this still need to be merged? If so I can do it if you resolve the conflict.

Clownacy:
These recipes did confuse me when adding my own libretro core: it's unclear what files actually need adding to, and which files are just vestigial leftovers that could be ignored.

DAR3LITESQUAD (approved):

PR #1610, opened 2022-01-11
picodrive: Add .pco to the list of supported extensions

phcoder:


phcoder:
https://github.com/libretro/libretro-super/pull/1610

PR #1646, opened 2022-04-10
Info-file updates

keithbowes:
Mostly related to versions, but some other information too (like taking
the parts about NstDatabase.xml out of the Nestopia info file)


fpscan:
Could you also send your changes to: https://github.com/libretro/libretro-core-info

PR #1652, opened 2022-05-18
Delete fbalpha_libretro.info

barbudreadmon:
i have only one word for the person who added back that core : NO


barbudreadmon:
>If the core shouldn't be included, then remove it from the build bot.

I can only agree with that, it was added by mistake, hence why the info file was left missing on purpose so that the core wouldn't appear on the online updater.
I suppose i should have totally deleted the old repo when we changed name...

ReenigneArcher:
> > If the core shouldn't be included, then remove it from the build bot.
> 
> I can only agree with that, it was added by mistake, hence why the info file was left missing on purpose so that the core wouldn't appear on the online updater. I suppose i should have totally deleted the old repo when we changed name...

I don't know how the cores are actually added to the build bot. I only discovered the missing info file as I created a script that parses the core files and info files (for my front end game server), and there was a mismatch.

The documentation about cores is also lacking a lot. It's very difficult to find any information about them.

RobLoach:
The docs could very much improve:
- https://github.com/libretro/docs/tree/master/docs/library
- https://docs.libretro.com/

PR #1654, opened 2022-05-30
Add ep128emu_libretro.info

zoltanvb:
This PR adds an info file for the new [ep128emu core](https://github.com/zoltanvb/ep128emu-core).


LibretroAdmin:
Hi there, if you'd like to get your core up on the buildbot, let us know. We could have some talks.

PR #1701, opened 2022-12-20
Add ChimeraSNES.

jamsilva:
This is a new core based on snes9x, made primarily for slower systems.

### DIFF
--- a/recipes/android/cores-android
+++ b/recipes/android/cores-android
@@ -96,6 +96,7 @@ smsplus libretro-smsplus https://github.com/libretro/smsplus-gx.git master YES G
 snes9x libretro-snes9x https://github.com/libretro/snes9x.git master YES GENERIC_JNI Makefile libretro/jni
 snes9x2005 libretro-snes9x2005 https://github.com/libretro/snes9x2005.git master YES GENERIC_JNI Makefile jni
 snes9x2005_plus libretro-snes9x2005_plus https://github.com/libretro/snes9x2005.git master YES GENERIC_JNI Makefile jni USE_BLARGG_APU=1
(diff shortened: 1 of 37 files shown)


recipes/android/cores-android
```diff
@@ -96,6 +96,7 @@ smsplus libretro-smsplus https://github.com/libretro/smsplus-gx.git master YES G
 snes9x libretro-snes9x https://github.com/libretro/snes9x.git master YES GENERIC_JNI Makefile libretro/jni
 snes9x2005 libretro-snes9x2005 https://github.com/libretro/snes9x2005.git master YES GENERIC_JNI Makefile jni
 snes9x2005_plus libretro-snes9x2005_plus https://github.com/libretro/snes9x2005.git master YES GENERIC_JNI Makefile jni USE_BLARGG_APU=1
+chimerasnes libretro-chimerasnes https://github.com/jamsilva/chimerasnes.git master YES GENERIC_JNI Makefile jni
```
RobLoach, line 99: I could be wrong, but I think these recipes are no longer used, and the gitlab CI is prioritized now :thinking:

jamsilva: Hmm... I just added ChimeraSNES to all the places where snes9x2005 also exists because I based the libretro support on it.

hizzlekizzle: We indeed no longer use the libretro-super scripts for our automated builds, but end-users frequently use them to compile their own cores, so there's still good reason to add these recipes. It just won't automatically get them added to RetroArch's online updater, for example

LibretroAdmin: Yeah, contributions to libretro-super are still fine and welcomed.

jamsilva: Is there anything I need to do additionally before this can be merged? I see that another new core got merged recently but not this one...

PR #1725, opened 2023-05-08
Remove unnecessary BIOS from freej2me

Jarvik7:
Reference: https://github.com/hex007/freej2me

Only -lr is needed for the libretro core


RobLoach (approved):
Thanks!

RobLoach (approved):

PR #1753, opened 2023-09-05
database: Build the Vircon32 RDB

RobLoach:
Related to #1752


gouchi:
When it is merged we can review [thumbnails](https://github.com/libretro-thumbnails/libretro-thumbnails/issues/131).

PR #1764, opened 2023-10-30
Add Double Cherry GB core

TimOelrichs:
Added my core to rules and recipes. Hope i did this right.


hizzlekizzle:
the actual recipe changes look right, but there seems to be a few erroneously added files in there and a number of files marked as complete replacements but with no actual changes.

PR #1773, opened 2023-12-08
db: Build the Spectravideo database

RobLoach:


WizzardSK:
Please add also "Commodore - PET" database to build

PR #1790, opened 2024-02-07
Update beetle psx hw savestate features to match non-hw

warmenhoven:


hizzlekizzle:
I believe in this case the savestate features were purposely incorrect to avoid people trying to use netplay/runahead with the hardware renderers.

warmenhoven:
Should this be reverted then? I've been thinking about this since you left your comment, and decided I don't think I know enough to know what the right thing to do is. I would have expected that the visual glitches of netplay/runahead would mostly be minor and short-lived, but I'm sure there's something I'm not thinking of/aware of.

hizzlekizzle:
I'm not an expert on it, but I don't think runahead/netplay/preemptive frames act like people would expect when hardware renderers are involved. Like you, I don't know whether this is enough of a problem to preclude using them altogether. If so, perhaps a better solution would be to disable them when the info file has a hardware renderer listed instead of piggybacking on the savestates like this was doing.

Do you have any thoughts, @Tatsuya79 ?

Tatsuya79:
Software runahead is working fine, HW is working too in single instance mode I've just tried it.
It's crashing if you switch fullscreen/windowed but it seems ok that aside..

hizzlekizzle:
k, in that case, I think we're probably fine here as-is. false alarm :)

warmenhoven:
Better to have a fire drill than ignore a real problem.

PR #1797, opened 2024-02-21
Clean up .info files

Jarvik7:
This is primarily regarding the hashes for bioses.
Currently the hashes are all smashed into a single notes line, which is hard to parse programmatically, and even harder to visually confirm that it is complete (many bios are missing hashes)

This is the first step of an intended cleanup with the end goal of having hashes for all bios, and separately a programmatically generated .dat file for each platform to enable users to easily gather the required bios (and only the required bios) for their platform.

Roadmap is as follows but there should be nothing stopping the immediate commit of this PR.

Roadmap:
1) Move hashes from notes line into separate firmware#_md5 or firmware#_sha1 lines for each bios definition
2) Create python script to generate master system.dat from folder of .info files
3) Add missing bios hashes and switch to sha1
4) Create additional python script to generate a platform-specific system.dat, e.g. dat for bios needed on cores for PSV retroarch build. This will use platform recipes to select the .info files to use


zoltanvb:
It does look more readable in the file, but this change with current RetroArch code will make the checksum values disappear from Information / Core Information after loading a core. Firmware checksums are not validated (at least not by RA, anyway, as far as I could check from https://github.com/libretro/RetroArch/blob/master/core_info.c ), but still this is a loss of information for the end user.

Note for anyone testing this, remove the core_info.cache file after updating .info files.

RobLoach:
Always a fan of more cleanup in these info files.

> Move hashes from notes line into separate firmware#_md5 or firmware#_sha1 lines for each bios definition

Is this change supported by RetroArch currently? I haven't seen many .info files using this new format.

> Can you also do the samething here: https://github.com/libretro/libretro-core-info

That's just a mirror of this, so likely best to handle it here first.

RobLoach:
Looks like there's some conflicts now. Always good to clean these up.

dist/info/bsnes_hd_beta_libretro.info
```diff
@@ -15,64 +15,86 @@ systemid = "super_nes"
...
 firmware17_opt = "true"
+firmware17_md5 = "b15ddb15721c657d82c5bab6db982ee9"
+
 firmware18_desc = "SGB2.sfc (SGB Boot Image)"
 firmware18_path = "SGB2.sfc"
 firmware18_opt = "true"
-notes = "[1] bsnes uses split ROMS for special chip games.|[*] Notable DSP1/DSP1B Games: Super Mario Kart,  Pilotwings|[*] Notable DSP2 Games: Dungeon Master|[*] Notable DSP3 Games: SD Gundam GX|[*] Notable DSP4 Games: Top Gear 3000|[*] Notable Cx4 Games: Mega Man X2,  Mega Man X3|[2] SGB Emulation needs No-Intro Super Game Boy ROMs renamed to SGB1.sfc or SGB2.sfc|(!) SGB1.sfc (md5): b15ddb15721c657d82c5bab6db982ee9|(!) SGB2.sfc (md5): 8ecd73eb4edf7ed7e81aef1be80031d5"
+firmware18_md5 = "8ecd73eb4edf7ed7e81aef1be80031d5"
+
+notes = "[1] bsnes uses split ROMS for special chip games.|[*] Notable DSP1/DSP1B Games: Super Mario Kart,  Pilotwings|[*] Notable DSP2 Games: Dungeon Master|[*] Notable DSP3 Games: SD Gundam GX|[*] Notable DSP4 Games: Top Gear 3000|[*] Notable Cx4 Games: Mega Man X2,  Mega Man X3|[2] SGB Emulation needs No-Intro Super Game Boy ROMs renamed to SGB1.sfc or SGB2.sfc
```
RobLoach, line 97: Agreed, looks like a few cases got split.

RobLoach (changes requested):

Issue #1801, opened 2024-03-16
melonDS DS missing, cannot be built


DJtheMan2101:
It looks like the [melonDS DS](https://github.com/JesseTG/melonds-ds) core is missing from [core-rules.sh](https://github.com/libretro/libretro-super/blob/master/rules.d/core-rules.sh), so it cannot be built from libretro-super. It has to be built from its source repo. For some reason, it never got added here even after it was released on the buildbot. Having it here would be pretty convenient, since all the cores could be built from one place.

RobLoach:
Thanks. This can be closed.

PR #1802, opened 2024-03-24
Rule updates: add geolith, update emuscv URL

zoltanvb:


zoltanvb:
+ added MelonDS DS (thanks for reporting).

Closes #1801

PR #1810, opened 2024-04-30
pico8: Add PICO-8 database for Retro8 and FAKE-08

RobLoach:


RobLoach:
@fpscan Thanks! Can we get this one too? https://github.com/libretro/libretro-super/pull/1808

PR #1813, opened 2024-05-19
Delete 4do info file

gouchi:
It has been renamed to Opera since this [commit](https://github.com/libretro/libretro-super/pull/1230/).

So we should remove it as it is not provided anymore on the buildbot.

@trapexit Do we keep it for compatibility reason ?

Thank you.


trapexit:
I don't believe so.

hizzlekizzle:
We have a few others in the same boat, such as a few mupen flavors that no longer exist.

fpscan:
So this one good to remove?

PR #1816, opened 2024-05-30
Add 4DO .info file

RobLoach:
Do we need this anymore? Was this core renamed or something?


gouchi:
I think we should remove it as it has been renamed [Opera](https://github.com/libretro/libretro-super/blob/master/dist/info/opera_libretro.info) ?

ping @trapexit

trapexit:
https://github.com/libretro/libretro-super/pull/1813

RobLoach:
Thanks for the clarification :+1:

PR #1823, opened 2024-07-07
Update swanstation_libretro.info

DarthMew:
Rewind does not work with SwanStation, so it's better to just disable it.


miguedevel:
I've been using rewind perfectly fine with this core for at least 6 months and today, after trying out latest nightly it suddenly didn't only to find out that it's because this file was changed.

PR #1825, opened 2024-07-10
Revert "Update swanstation_libretro.info"

miguedevel:
As I commented on libretro/libretro-super#1823 rewind works perfectly fine on the Swanstation core.

### DIFF
--- a/dist/info/swanstation_libretro.info
+++ b/dist/info/swanstation_libretro.info
@@ -20,7 +20,7 @@ hw_render = "true"
 required_hw_api = "OpenGL >= 3.0 | OpenGL Core >= 3.1 | Vulkan >= 1.0 | Direct3D >= 11.0"
 is_experimental = "false"
 savestate = "true"
-savestate_features = "basic"
+savestate_features = "serialized"
 input_descriptors = "true"
 disk_control = "true"
 


DarthMew:
I think your definition of "perfectly fine" is way too optimistic here.
Anyway, you can bypass the restriction by enabling the "Bypass Core Info Save States Features" option inside RetroArch's Core submenu, but I'm not dealing with a feature that I know (currently) doesn't work correctly with SwanStation's JIT, so please close this PR.

miguedevel:
OK so what you're saying it doesn't work when using recompiler mode but it works fine on interpreter mode? I even switched to recompiler mode now and seems to work fine.
Is this bug only on specific platforms?
Can you explain why I'm being too optimistic with saying that it works fine?

DarthMew:
On systems that are easily capable of handling rewind with heavy cores like Beetle PSX and even Beetle Saturn , it rewinds SwanStation with such reduced performance that it causes noticable audio crackling and slow rewind speed (<40fps with JIT enabled vs a smooth 60fps with the interpreter, both with Rewind Frames set to 1).

As I've said before, people can bypass the restriction easily enough, but unless someone can assist in making sure rewind works allround with one of the key features of SwanStation (its JIT), I find it better to have it disabled as a precaution.

miguedevel:
These are my thoughts on this:
1. Out of the box, Rewind is disabled and the Rewind option clearly describes how it has a performance impact.
2.  Just as easy as it is to bypass savestate lock with an option manually by the user, so is disabling rewind.
3.  The support requests will change from "rewind is slow" to "hey rewind used to work why did it stopped working" (like me!)
4. If the slow rewind with JIT/recompiler is an emulator core bug, why is there no open issue on the repo for it (with reproducible steps) to give it more visibility so that maybe someone can tackle it.

hizzlekizzle:
You make some reasonable arguments, for sure :)

From the support perspective, sure, we have a brief period of recalibration as people who used rewind with the core in the past are surprised about losing it, but once that transition period ends, we're done. OTOH, if we leave it in a state that's suboptimal on an ongoing basis, those support questions never end.

Savestate shenanigans are generally poor-performing when it comes to dynarecs. I believe upstream disables the dynarecs whenever they do runahead, so we would probably need some mechanism for doing the same thing for rewind.

fpscan:
Alright, at this point it's good to close this PR.

PR #1828, opened 2024-08-07
Use the open source ps2dev compilers for Sony PS2

danboid:
Does anyone still use the ancient Sony PS2 SDK?

I had to change the definition of gcc and g++ for the PS2 before I could get libretro to build with the ps2dev toolchain:

https://github.com/ps2dev/ps2dev


zoltanvb:
PS2 compilations (at least on the buildbot) are done using other definitions. In any case, I guess this could be merged, in case someone still uses it from here.

fjtrujy:
Fell free to merge this PR

PR #1846, opened 2024-10-03
Add the new Holani core

LLeny:
[Holani](https://github.com/LLeny/holani-retro) is a new cycle stepped Atari Lynx emulator made in rust. It is not a Handy port.
Its end goal is to get closer to the Lynx hardware to provide a more accurate emulation.
The current Gitlab CI should create the following targets:
* linux-x86_64
* windows-x86_64
* osx-x86_64
* osx-aarch64
* android-arm64-v8a
* android-armeabi-v7a

However, only the linux and windows cores have been tested at this point.


hizzlekizzle:
Sounds exciting!

PR #1869, opened 2024-12-27
Update display name in atari800 info file

garbear:
## Description

This PR updates the display name for the atari800 emulator. It was shortened to "Atari - 5200 (Atari800)" when display names were converted from long strings to `platform (name)` in commit: https://github.com/libretro/libretro-super/commit/bd7bc168d3da8497a30aa63d22abe4fbe0c46e9a#diff-14dad692fb490fa8582eadfc12c41eced0ee17849104e388c84f0ac55230bd0eL1-R1

```patch
-Atari 8-bit computer systems and 5200 (Atari800)
+Atari - 5200 (Atari800)
```

However, according to the Atari 800 "Features" page (https://atari800.github.io/features.html), atari800 supports "Atari 400, 800, 600 XL, 800XL, 130XE, 5200 Games System".

So a better name for the atari800 core would be:

* `Atari - 400/800/600XL/800XL/130XE/5200 (Atari800)`

## Screenshots

Before:

Screenshot 2024-12-26 at 6 35 14 PM

After:

Screenshot 2024-12-26 at 6 21 02 PM

### DIFF
--- a/dist/info/atari800_libretro.info
+++ b/dist/info/atari800_libretro.info
@@ -1,5 +1,5 @@
 # Software Information
-display_name = "Atari - 5200 (Atari800)"
+display_name = "Atari - 400/800/600XL/800XL/130XE/5200 (Atari800)"
 authors = "Petr Stehlik"
 supported_extensions = "xfd|atr|dcm|cas|bin|a52|zip|atx|car|rom|com|xex|m3u"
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
As long as the core is set up to actually load all of those, it seems like a slam dunk to me.

PR #1870, opened 2024-12-27
Update Beetle Wonderswan name

garbear:
## Description

This PR fixes the name of Beetle Wonderswan. The core was forked from Cygne when it was integrated into Mednafen.

### DIFF
--- a/dist/info/mednafen_wswan_libretro.info
+++ b/dist/info/mednafen_wswan_libretro.info
@@ -1,5 +1,5 @@
 # Software Information
-display_name = "Bandai - WonderSwan/Color (Beetle Cygne)"
+display_name = "Bandai - WonderSwan/Color (Beetle Wonderswan)"
 authors = "Dox|Mednafen Team"
 supported_extensions = "ws|wsc|pc2|pcv2"
 corename = "Beetle WonderSwan"


fpscan:
@hizzlekizzle

hizzlekizzle:
I don't feel strongly about this either way, but I believe it was changed back around the same time we changed the various Snes9x forks back to Snes9x names--instead of, for example, CatSFC.

Similarly, we call the bsnes-derived SNES core 'Beetle bsnes'.

garbear:
I don't think it's ever been changed? The initial commit 13 years ago was:

```
display_name = "WonderSwan/Color (Mednafen Cygne)"
supported_extensions = "ws|wsc"
corename = "Mednafen WonderSwan"
manufacturer = "Bandai"
systemname = "WonderSwan/Color"
```

We could argue for consistency, in which case it would be probably be wonderswan. The file is named `mednafen_wswan_libretro.info`, so even mednefen was producing the wonderswan name for the core back 13 years ago.

hizzlekizzle:
AFAIK, this won't change anything like config/save file locations, which are all based on the core's internal name. If that's indeed the case, I'm fine with changing it /shrug

We should probably ask around with some other folks to make sure first, though.

garbear:
Correct, this is intended to be a purely visual change. For example, in Kodi, the add-on name will be shown differently. But there's no other affect.

garbear:
@LibretroAdmin In addition to being a superhero vanquisher of libretro bugs, can you check out https://github.com/libretro/xrick-libretro/pull/23? It addresses a misuse of the libretro API. ❤️ for all you do

PR #1871, opened 2025-01-01
bios: Update ecwolf.pk3

RobLoach:
This updates ecwolf.pk3 to what's listed in libretro-database's System.dat.

@hunterk Do you know if this is wrong/right?


hizzlekizzle:
That's a good question. Mine doesn't match either one lol

RobLoach:
Might as well bring it in then? :rofl:

hizzlekizzle:
Sure, it doesn't make it any worse, AFAICT, and at least it makes our documentation consistent :P

PR #1875, opened 2025-01-11
Update pcsx2_libretro.info

bslenul:
Adds a "GameIndex.yaml" check and a small description:

image

I marked it as required, although this isn't entirely true the file is soooo important that I feel like it would be very misleading to mark it as optional.

**edit:** My English isn't the best, if you have a better description please let me know, I'll update the PR :p

### DIFF
--- a/dist/info/pcsx2_libretro.info
+++ b/dist/info/pcsx2_libretro.info
@@ -25,10 +25,13 @@ savestate_features = "basic"
 memory_descriptors = "false"
 
 # Firmware / BIOS
-firmware_count = 1
+firmware_count = 2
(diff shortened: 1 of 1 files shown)


fpscan:
@hizzlekizzle maybe hunter can help us for proofreading.

hizzlekizzle:
Seems fine to me. I would probably make these very small changes:

> The 'GameIndex.yaml' file can be downloaded from the Online Updater.|[^] It contains fixes for a lot of games and is mandatory|[^] for best compatibility. Some games may not even boot without it.

bslenul:
Thank you! Description updated 👍

PR #1878, opened 2025-01-13
Pick a smaller set of platforms for DICE.

mittonk:
linux-x64 and linux-i686 are working, let's start with those.

osx-arm64 and windows-x64 build on my local test systems, so those should be doable in a while.

The rest I'll leave for a while.


mittonk:
(Starting with the .gitlab-ci.yml end of things instead.)

PR #1879, opened 2025-01-16
More platforms for DICE --- that'll help the PSP and WiiU builds.

mittonk:
More plaforms for DICE --- that'll help the PSP and WiiU builds.


zoltanvb:
Hi, there is no harm in adding new cores to recipes, but current buildbot is not using those. (If you found it in a piece of documentation, I can fix that as well.)

Buildbot integration needs 2 things:
- a mirror of the repository from github to the libretro gitlab server (this needs to be set up by libretro admins)
- a .gitlab-ci.yml file in the repository root, which describes the CI jobs - there is no guide for that, but based on the recipe changes, it should be an easy update from a working example

mittonk:
Thanks for the pointers --- I did find and fix issues in `.gitlab-ci.yml` in my repo that were part of the problem I was trying to solve here.

I didn't find any stale documentation, was just grepping around.

Thanks!

PR #1890, opened 2025-04-01
lr-dice earns a version bump with DIP switch support and core_options_v2.

mittonk:


mittonk:
Going with a bigger branch instead.

PR #1892, opened 2025-04-18
Fix ROM Cleaner extension.

zoltanvb:
Original PR #1889 added it without extension?

@JesseTG can comment if that was intentional.


JesseTG:
Whoops. Thanks for this.

PR #1905, opened 2025-06-03
Fix opengles flag

cscarnell98:
The flag is enable-opengles not enable-gles ASFAIK!


hizzlekizzle:
yes, you're correct. Thanks for the fix!

PR #1907, opened 2025-06-15
Add uxn core.

iyzsong:
Hello, this is a core for the [Uxn](https://100r.co/site/uxn.html).
Require zig to build, only tested for Linux.


hizzlekizzle:
Looks cool!

PR #1915, opened 2025-08-02
Update gam4980_libretro.info

iyzsong:


gouchi:
@fpscan This should be good to apply but it seems there is ?

> This branch has conflicts that must be resolved
> Changes can be cleanly merged.

PR #1968, opened 2026-03-21
VICE + PUAE version update

sonninnos:
Just bumping VICE version to current 3.10 and PUAE to 5.3.1.

For some reason after pulling the latest there were "unstaged changes" in `trident_libretro.info` which I absolutely did not touch in any way..


sonninnos:
Something needs fixing?

PR #1969, opened 2026-03-21
scummvm: Update database to scan only .scummvm files

RobLoach:
Been working on the ScummVM database to only read through `.scummvm` files, built out from the list when running `scummvm --list-all-games`.

https://github.com/libretro/libretro-database/pull/1647

The automated scan was always pretty buggy, and this will ensure there's always something that actually works.

### DIFF
--- a/dist/info/scummvm_libretro.info
+++ b/dist/info/scummvm_libretro.info
@@ -1,7 +1,7 @@
 # Software Information
 display_name = "ScummVM"
 authors = "SCUMMVMdev"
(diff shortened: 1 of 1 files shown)


cscd98:
This has the side affect of no longer being able to find scummvm files, such as MONKEY.000 to load the game. The only alternative is to disable RA setting filter by known extensions?

RobLoach:
Correct. Having all the file extensions listed was unmanageable, and it didn't work for games which didn't have a file in the root.

Unfortunate, but likely the best workaround for now.

PR #1980, opened 2026-06-02
Add Lost Level Archive to Database

Zoomll:
Added Lost Level Archive to the database build script.

A PR in [libretro/libretro-database](https://github.com/libretro/libretro-database) reflecting this change will be created soon.

### DIFF
--- a/libretro-build-database.sh
+++ b/libretro-build-database.sh
@@ -230,6 +230,11 @@ build_libretro_database() {
 			COMMAND+=' "${LIBRETRODATABASE_META_DAT_DIR}/no-intro/${1}.dat"'
 		fi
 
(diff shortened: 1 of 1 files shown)


Zoomll:
libretro-database PR: https://github.com/libretro/libretro-database/pull/1670

zoltanvb:
Hi, thank you for adding this. One question to whoever feels like they are the "concept owner" of the Libretro DB: as I remember from my previous investigations, the order of adding the databases in the build script is also an implicit priority order (see https://github.com/libretro/libretro-database#sources), in reverse. So, currently in case of any conflict, lost-level-archive data will override/hide the no-intro/redump data - is this desired?

PR #1984, opened 2026-06-15
Add Eternal Computer core

iyzsong:
Hello, here is a core for the Eternal Computer, which is a simple one instruction virtual machine.
It can run linux and doom, but no sound, only keyboard.

See:
  https://www.ioccc.org/2025/cable/index.html
  https://www.eternal-software.org
  https://github.com/adriancable/eternal


RobLoach:
You'll want to update to the new CMakeLists.txt to get this workign on the buildbot:
https://github.com/libretro/skeletor

PR #2000, opened 2026-07-03
by request of upstream

warmenhoven:


hizzlekizzle:
With the disambiguation, would it make sense to take it off the "experimental" list while we're at it?

warmenhoven:
🤷 I don't have a strong opinion on it. I guess I lean mildly toward no, leave it as experimental, mostly out of entropy, but also because there are still a few games that just don't run.

PR #2013, opened 2026-07-13
Add the yaps2 libretro core (info file + linux x64 recipe, experimental)

WizzardSK:
Adds **yaps2**, a libretro core for the yaps2 PlayStation 2 emulator (a PCSX2 fork aimed at ARM handhelds: working arm64 recompilers, Adreno fixes). Marked `is_experimental = "true"`.

The libretro frontend was merged into `yaps2/yaps2` main as yaps2/yaps2#4 (`pcsx2-libretro/`, built with `-DENABLE_LIBRETRO=ON`, producing `yaps2_libretro.so`), so this PR adds both the info file and a build recipe pointing at upstream main.

Tested on a Snapdragon 720G handheld (Adreno 618, Turnip): GT3 boots and renders, save states, `.m3u` disc swapping and the core options all work.

Every info field was checked against `pcsx2-libretro/Main.cpp` rather than copied from another PS2 core:

- `hw_render = true`, `required_hw_api = "Vulkan >= 1.1"` — unlike the other PS2 cores, this one genuinely shares the frontend's Vulkan context: it takes the VkInstance/VkPhysicalDevice from `SET_HW_RENDER_CONTEXT_NEGOTIATION_INTERFACE` and hands finished frames back via `set_image`. There is no software/GL fallback toward the frontend, so a Vulkan driver is mandatory.
- `disk_control = true` — `SET_DISK_CONTROL_EXT_INTERFACE`, `.m3u` playlists and tray swap.
- `core_options_version = 2.0` — `SET_CORE_OPTIONS_V2`.
- `supports_no_game = false`, `savestate = true` (`savestate_features = basic`).
- `cheats = false`, `memory_descriptors = false`, `input_descriptors = false` — none of those interfaces are registered (`retro_cheat_set` is an empty stub, `retro_get_memory_data` returns null). The `yaps2_cheats` core option is the emulator's own pnach system, not libretro cheats.
- Firmware: `<system>/pcsx2/bios` and `<system>/pcsx2/resources`, both mandatory, following the folder mapping in `Main.cpp`.

`systemid`/`systemname` match the other PS2 cores in `dist/info`.

The recipe is the one that builds the core here (`-DENABLE_LIBRETRO=ON -DENABLE_QT_UI=OFF -DENABLE_TESTS=OFF`); I have not been able to try it on the buildbot image, so if its dependency set needs anything else, tell me and I will adjust.

### DIFF
--- a/dist/info/yaps2_libretro.info
+++ b/dist/info/yaps2_libretro.info
@@ -1,0 +1,44 @@
+# Software Information
+display_name = "Sony - PlayStation 2 (yaps2)"
+authors = "PCSX2 Team|bmdhacks|WizzardSK"
+supported_extensions = "elf|iso|ciso|chd|cso|zso|bin|mdf|nrg|dump|gz|img|irx|m3u"
+corename = "yaps2"
+categories = "Emulator"
+license = "GPLv3"
+permissions = ""
+display_version = "Git"
+
+# Hardware Information
+manufacturer = "Sony"
+systemname = "Sony PlayStation 2"
+systemid = "playstation2"
+
+# Libretro Features
+database = "Sony - PlayStation 2"
+supports_no_game = "false"
+savestate = "true"
+savestate_features = "basic"
+cheats = "false"
+input_descriptors = "false"
+memory_descriptors = "false"
+libretro_saves = "false"
+core_options = "true"
+core_options_version = "2.0"
+load_subsystem = "false"
+hw_render = "true"
+required_hw_api = "Vulkan >= 1.1"
+needs_fullpath = "true"
+disk_control = "true"
+is_experimental = "true"
+
+# Firmware / BIOS
+firmware_count = 2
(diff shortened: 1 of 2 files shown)


WizzardSK:
Following up on this: yaps2 hasn't shown up in the `nightly/linux/x86_64` listing yet, and since the `linux/aarch64` nightlies are driven by per-core GitLab CI rather than these recipes, I've added a `.gitlab-ci.yml` upstream (modeled on flycast's, using the `linux-cmake.yml` ci-templates): yaps2/yaps2#9

It defines `libretro-build-linux-x64` and `libretro-build-linux-aarch64` jobs. The buildbot images ship neither Clang nor the pinned third-party libs the core needs, so the job's `before_script` installs clang-17/lld-17 from apt.llvm.org and builds the deps with the repo's own script — I can't exercise your runners from here, so happy to iterate if that needs adjusting for your images.

Could you enable the core on git.libretro.com (mirror + CI) for linux x64 and aarch64 when you get a chance? Thanks!

PR #2028, opened 2026-07-26
Rule updates: fix tic80 build binary dir, update scummvm source, add scummvm WIP engines

zoltanvb:


zoltanvb:
+ uae4arm makefile location corrected.

PR #2037, opened 2026-07-30
Add Cemu (Wii U) core rule and info file

WizzardSK:
There is no Wii U core on the buildbot at the moment. [Cemu](https://github.com/cemu-project/Cemu) has a libretro port that builds and runs on Linux, so this registers it: a core rule plus a `.info` file.

### What was verified

I built the core natively on aarch64 (GCC 15) and ran it in RetroArch on an Adreno 618 / Turnip machine. New Super Mario Bros. U boots and renders on the Vulkan path at ~33 FPS on the title screen.

Every field in the info file was read off the core, not guessed:

| field | where it comes from |
| --- | --- |
| `supported_extensions`, `needs_fullpath`, `display_version` | `retro_get_system_info` |
| `hw_render = "true"` | core requests a HW render context (Vulkan or GL 4.5 core profile) |
| `input_descriptors = "true"` | `RETRO_ENVIRONMENT_SET_INPUT_DESCRIPTORS` |
| `libretro_saves = "true"` | MLC storage is created under `GET_SAVE_DIRECTORY` |
| `savestate = "false"` | `retro_serialize_size` returns 0 by design - Cemu has no state serialization |
| `cheats = "false"` | `retro_cheat_set` is an empty stub |
| `memory_descriptors = "false"` | the core never calls `SET_MEMORY_MAPS` |

Marked `is_experimental = "true"`.

`build_extradir` is `../bin/` because Cemu's CMake writes `cemu_libretro.so` to `<source>/bin` rather than into the build directory.

### Two things worth flagging

- **The URL points at my fork.** Upstream `danprice142/Cemu-Libretro` has been dead since January and is ~119 commits behind; the libretro work (Vulkan HW render via context negotiation, DRC screen layouts, core options) lives on my branch. Happy to retarget if the core gets a home under the libretro organisation.
- **There is no project for cemu on git.libretro.com** (`/api/v4/projects?search=cemu` returns `[]`), so nothing will actually build until a mirror is registered. The repository has a `.gitlab-ci.yml` with a `libretro-build-linux-aarch64` job ready and its default branch is set to the branch carrying it - but that job has not been run yet, since I have no way to trigger a pipeline.

Builds are heavy: vcpkg compiles roughly 30 ports from source, Boost included.

### DIFF
--- a/dist/info/cemu_libretro.info
+++ b/dist/info/cemu_libretro.info
@@ -1,0 +1,35 @@
+# Software Information
+display_name = "Nintendo - Wii U (Cemu)"
+authors = "Cemu Project|WizzardSK"
+supported_extensions = "wud|wux|wua|iso|rpx|elf"
+corename = "Cemu"
+categories = "Emulator"
+license = "MPL-2.0"
+permissions = ""
+display_version = "2.6"
+
+# Hardware Information
+manufacturer = "Nintendo"
+systemname = "Wii U"
+systemid = "wiiu"
+database = "Nintendo - Wii U"
+
+# Libretro Features
+supports_no_game = "false"
+hw_render = "true"
+needs_fullpath = "true"
+disk_control = "false"
+savestate = "false"
+cheats = "false"
+input_descriptors = "true"
+memory_descriptors = "false"
+libretro_saves = "true"
+core_options = "true"
+load_subsystem = "false"
+is_experimental = "true"
+
+# Firmware / BIOS
+firmware_count = 0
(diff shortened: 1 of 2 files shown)


AloysHF:
@WizzardSK 
Thank you for your contribution, but I think you should try to focus on a specific optimization point. I have no problem with using AI—everyone relies on it nowadays—but I believe you should concentrate on a specific task rather than spreading your energy across every aspect. The buildbot registration for these RetroArch cores should be completed by the corresponding developers; only they will have enough bandwidth for the ongoing maintenance work.

PR #2059, opened 2026-08-07
info: BlastEm supports more than the Mega Drive

WizzardSK:
`blastem_libretro.info` still describes the core as a Mega Drive-only emulator, but the core has advertised the full Sega line-up for a while now:

```c
info->valid_extensions = "md|gen|smd|32x|sms|gg|sg|sg1|sc|sc3|sf7|col|cue|toc|iso|vgm|vgz|flac|wav|bin|rom|gz";
info->need_fullpath = 1;
```

Because RetroArch filters the file browser by the info file rather than by the core, every Master System, Game Gear, SG-1000, 32X and CD image is hidden once BlastEm is the selected core, and the core looks like it only works with Genesis content. This was reported by a user who had updated the core and could not work out why nothing else would load.

Changes, all of them matching what the core actually reports:

- `supported_extensions` — the core's own list.
- `needs_fullpath` — `true`, the core sets `need_fullpath = 1`.
- `core_options` / `core_options_version` — `true` / `0`, the core registers `blastem_system_type` through `RETRO_ENVIRONMENT_SET_VARIABLES`.
- `database` — adds 32X, Game Gear, Master System, Mega-CD, PICO and SG-1000 alongside the Mega Drive.
- `display_name` / `systemname` / `description` — reflect the supported systems.

### DIFF
--- a/dist/info/blastem_libretro.info
+++ b/dist/info/blastem_libretro.info
@@ -1,7 +1,7 @@
 # Software Information
 display_name = "Sega - MS/GG/MD/CD/32X (BlastEm)"
 authors = "Mike Pavone"
-supported_extensions = "md|gen|smd|68k|sgd|32x|sms|gg|sg|sg1|sc|sc3|sf7|col|cue|toc|iso|vgm|vgz|flac|wav|bin|rom|gz"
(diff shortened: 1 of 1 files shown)


WizzardSK:
Added `chd` to the extension list. It goes with libretro/blastem#55, which gives the core a CHD reader (vendored libchdr) - and until that merges the core reports a `.chd` as unsupported instead of misreading it, so the extension being listed ahead of time costs nothing.

Happy to drop that commit if you would rather this PR stayed to what the core does today.

PR #2064, opened 2026-08-10
Anarch fix plus potential argument resolution.

SparkyKuriette:
Anarch has Libretro-standard save support (at a measly 12 bytes per save), but no savestate support, so I, um, don't know why that its savestates are listed as "deterministic" within RetroArch, but its info file has "false" as its save support.\
\
Also, Anarch's creator, Miloslav "drummyfish" Číž, supports, for example, pædophilia and zoophilia, and is a racist (he frequently uses a slur against non-white people that starts with N), ableist (his wiki has an anti-mentally-disabled slur in its name) and transphobe (he thinks that trans women are "men"). The creator's history caused an argument between some people, and I got involved. See https://github.com/libretro/RetroArch/issues/19348 (marked as completed, by the way). This is an attempt at a compromise.\
\
His wiki (note the anti-mentally-disabled slur): https://git.coom.tech/drummyfish/less_retarded_wiki \
\
(Addendum 1, 2026-09-17: Fixed issue)


hizzlekizzle:
Good catches on the savestate stuff.

I'm a little concerned about the disclaimer implying that we *do* endorse the opinions of other core authors, though. Obviously none (that I'm aware of) are as unpleasant as this one, but we simply don't have that sort of relationship with most core authors to even know--as is/was the case here.

SparkyKuriette:
I see! Should I just remove the disclaimer then, but leave the edited save(state) info as-is?

hizzlekizzle:
eh, let's let this one simmer for a bit. I'll try to solicit some opinions from other contributors to see if they think it's more/less confusing with/without it.

SparkyKuriette:
All right. By the way, upstream/original Anarch has an optional mod for peer-to-peer multiplayer. I wonder if that would be feasible to add, assuming Anarch is kept.

RobLoach:
Thanks for taking action here. Sounds like this is still cooking a bit.

SparkyKuriette:
@hizzlekizzle, how is the, um, simmering going? It's been a month, and I'm only wanting to know how the progress is going.

PR #2065, opened 2026-08-12
Enable submodules for theodore core

Zlika:
Currently the theodore core includes a copy of some files from libretro-common but I would like to replace it by a submodule pointing to the libretro-common repository. If I understand correctly, I then need to create this "libretro_theodore_git_submodules" variable to make the build works.


fpscan:
@hizzlekizzle any input here?

PR #2072, opened 2026-08-20
virtualjaguar: update info for v3.4.0

JoeMatt:
Updates `dist/info/virtualjaguar_libretro.info` to match the [v3.4.0 release](https://github.com/libretro/virtualjaguar-libretro/releases/tag/v3.4.0) of `libretro/virtualjaguar-libretro`, tagged today. The file is copied verbatim from `dist/info/virtualjaguar_libretro.info` at that tag, so the two stay in sync.

### Changes

**`display_version`** — `v3.3.0` → `v3.4.0`.

**`supported_extensions`** — adds `chd`. The core now loads Jaguar CD discs from CHD in addition to CUE/BIN and CDI.

**`firmware_count`** — `2` → `4`. Adds the two cart boot ROM entries, both optional:

- `jagboot_m.rom` — overrides the embedded Model-M cart boot ROM
- `jagboot.rom` — used when the Cart BIOS Type option is set to Custom

**`notes`** — expanded from one line to five. The previous text said only that "an optional CD BIOS file" overrides the embedded copy, which did not tell users what to name it. The new notes enumerate every accepted filename and the subdirectories searched, and explain that images are identified **by checksum rather than filename** — so a "Missing" firmware row in the RetroArch UI is harmless if the user has the same image under any other accepted name. That mismatch was a recurring source of confusion.

**`description`** — mentions CHD support, and adds the constraint that CHD images must be produced by a chdman that writes CHSE session metadata (MAME 2026-08-15 / PR [15886](https://github.com/mamedev/mame/pull/15886) or later). Older CHDs found online lack it and are refused with a clear message; the core ships `jagcd` helper tools for converting.

### Notes

No BIOS files are required to use the core at all — the Jaguar boot ROM and CD BIOS images are embedded. Every firmware entry here is an optional override, and all four are marked `_opt = "true"`.


JoeMatt:
replaced by #2081

PR #2080, opened 2026-08-25
hatariB new Atari ST core

bbbradsmith:
info file for hatariB, a new core that I am preparing to submit
https://github.com/bbbradsmith/hatariB/

This info file is already present in the source repository, I don't know if there might be a way to automatically mirror it?
https://github.com/bbbradsmith/hatariB/blob/main/info/hatarib.info


hizzlekizzle:
no way to automatically mirror it, unfortunately

PR #2106, opened 2026-09-15
hatarib_libretro.info v1.0

bbbradsmith:
Final update for version 1.0 release.


dist/info/hatarib_libretro.info
```diff
@@ -1,12 +1,12 @@
 # Software
 display_name = "Atari - ST/STE/TT/Falcon (hatariB)"
 categories = "Emulator"
-authors = "Brad Smith|Hatari: Nicolas Pomarede et al."
-corename = "hatariB"
+authors = "Brad Smith|Hatari: Nicolas Pomarède et al."
```
RobLoach, line 4: Do you know if other .info files use UTF8 characters? While I would like to have them, I'm unsure if they're okay to use.

bbbradsmith: The other libretro hatari info has used the same for about 12 years: https://github.com/libretro/libretro-super/blob/3caa2318c8b47a838060335d29dbc0f4634a4fde/dist/info/hatari_libretro.info#L3

It also displays correctly in any version of Retroarch I have tried recently.

RobLoach: Thanks so much!

PR #2108, opened 2026-09-16
info: Add .info files from core-info

RobLoach:
These are a couple additions that were added to libretro-core-info, but missing from libretro-super.


RobLoach:
@WizzardSK These were missing from super, so here they are... If there are other changes in libretro-core-info you made, they'll go missing.

PR #2111, opened 2026-09-17
info: fill in the EmuSCV capability fields and description

MaaaX-EPOCH84:
Every capability field in emuscv_libretro.info was commented out with ???, and there was no description, so the Core Downloader showed nothing about the core. I am the core's author; each field below was checked against the source rather than guessed.

savestate = "true", savestate_features = "serialized" — retro_serialize_size() returns a real size, and rewind is in daily use on the distributions that ship the core.
needs_fullpath = "true" — the core reads the cartridge itself, including zip archives, and blocks extraction. Note that the core's own log line claims the opposite; the code is what I have followed.
input_descriptors = "true" — RETRO_ENVIRONMENT_SET_INPUT_DESCRIPTORS and SET_CONTROLLER_INFO are both set.
libretro_saves = "false" — cartridge RAM is deliberately not exposed to the frontend; the core writes its own save file.
cheats, memory_descriptors, hw_render, disk_control, load_subsystem — all false, no corresponding call exists.
display_name gains the core name to match the convention used elsewhere, and Toshiya Takeda is added as an author: the core is a port of his eSCV.


gouchi:
@fpscan Should be good to go ;)

